PR #2, opened 2016-10-06
Proposal for map2i and map2

raphael-proust:
Here's a quick implementation of `map2i` and `map2`. There are a few things that might need to change:
- Error management: the error when the number of columns or lines don't match should be similar to other functions with the same requirements (e.g., `add`).
- More functions: there is a collection of `map` functions (by row, by col, etc.) that should be available in `2` flavour.

In general, libraries that implement `map`-like and `map2`-like operations provide optimised versions up to a given number and then a generic one that is less efficient. Not sure you want a similar thing here.

### DIFF
--- a/lib/dense.ml
+++ b/lib/dense.ml
@@ -239,6 +239,15 @@ let mapi_by_col ?(d=0) f x =
 
 let map_by_col ?(d=0) f x = mapi_by_col ~d (fun _ y -> f y) x
 
+let map2i f x y =
+  if row_num x = row_num y && col_num x = col_num y then
+    let z = empty (row_num x) (col_num x) in
+    iteri (fun i j w -> Array2.unsafe_set z i j (f i j w)) x; y
+  else
+    raise (Invalid_argument "map2ij: dimensions do not match")
+
+let map2 f x y = mapi (fun _ _ w -> f w) x
+
 let filteri f x =
   let r = ref [||] in
   iteri (fun i j y ->
--- a/lib/dense.ml
+++ b/lib/dense.ml
@@ -242,11 +242,17 @@ let map_by_col ?(d=0) f x = mapi_by_col ~d (fun _ y -> f y) x
 let map2i f x y =
   if row_num x = row_num y && col_num x = col_num y then
     let z = empty (row_num x) (col_num x) in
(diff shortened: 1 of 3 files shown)


ryanrhymes:
Error management is a good idea. I will do that. Currently l leave this to the underlying GSL since it throws similar exceptions. 

could you double check map2i function? shouldn't it be "let map2 f x y = map2i ..." if my understanding correct.

lib/dense.ml
```diff
@@ -239,6 +239,15 @@ let mapi_by_col ?(d=0) f x =
 
 let map_by_col ?(d=0) f x = mapi_by_col ~d (fun _ y -> f y) x
 
+let map2i f x y =
+  if row_num x = row_num y && col_num x = col_num y then
+    let z = empty (row_num x) (col_num x) in
+    iteri (fun i j w -> Array2.unsafe_set z i j (f i j w)) x; y
```
sevenEng: If I understand the meaning of `map2i` correctly, maybe the type of `f` should be `val int -> int -> float -> float -> float` here, as it should take the values from both matrix `x` and `y`. Thus here it should be like something as
```ocaml
iteri (fun i j vx ->
  let vy = Array2.unsafe_get y i j in
  Array2.unsafe_set z i j (f i j vx vy)) x;
z
```
And the following `map2` function could be:
```ocaml
let map2 f x y = map2i (fun _ _ vx vy -> f vx vy) x y
```

sevenEng: If I understand the meaning of `map2i` correctly, maybe the type of `f` should be `val int -> int -> float -> float -> float` here, as it should take the values from both matrix `x` and `y`. Thus here it should be like something as
```ocaml
iteri (fun i j vx ->
  let vy = Array2.unsafe_get y i j in
  Array2.unsafe_set z i j (f i j vx vy)) x;
z
```
And the following `map2` function could be:
```ocaml
let map2 f x y = map2i (fun _ _ vx vy -> f vx vy) x y
```

raphael-proust:
Oh yeah, I wasn't very awake when I coded that and I didn't test it because it was a “let's start a conversation” kind of PR. I'll fix it.

raphael-proust:
Should be better now. I even added some tests.

PR #10, opened 2016-11-03
provisional opam support

struktured:
Edited some ```_oasis``` fields and added others so that ```oasis2opam``` accepted it. 

```opam pin -y add .``` now works like a charm to install Owl from the project root directory.

This is a little verbose as it was autogenerated- feel free to reject this PR and handcraft a simpler one if you prefer.

### CODE REVIEW

**Comment by ryanrhymes** [comment]
Thank you Carmelo, looks good to me.

File: Owl.install [hunk 1]
```diff
@@ -0,0 +1,6 @@
+etc: [
+  "setup.ml"
+  "setup.data"
+  "setup.log"
+  "_oasis_remove_.ml"
+]
```
hannesm: @struktured why is this needed?  shouldn't `install` below just call out to `ocaml setup.ml -install`?

_oasis_remove_.ml
```diff
@@ -0,0 +1,7 @@
+open Printf
+
+let () =
+  let dir = Sys.argv.(1) in
+  (try Sys.chdir dir
+   with _ -> eprintf "Cannot change directory to %s\n%!" dir);
+  exit (Sys.command "ocaml setup.ml -uninstall")
```
hannesm: I fail to see the need for this as well.. `ocamlfind remove owl` should do the trick!?

hannesm:
I put some comments inline... unfortunately I don't have a system to actually test this (since it seems to need a `gsl` library which is newer than what I have available (I don't have any `gsl/gsl_spmatrix.h` on my system)...

PR #13, opened 2016-12-13
Add a 'lacaml' opam dependency to match _oasis

yallop:


ryanrhymes:
@yallop cool, many thanks :)

PR #15, opened 2017-01-19
Sparse LDA 

bencatterall:
Inital implementation of the Sparse LDA algorithm. 

More work is needed to realise the full performance benefits. This includes using an alias table and a more efficient structure for storing non-zero probability factors.


ryanrhymes:
Thanks Ben. Well done 👍

PR #41, opened 2017-05-25
Added a test runner and Travis CI

superbobry:
This way we can make sure the PRs don't break anything (covered by tests).


unittest/unit_stats.ml
```diff
@@ -21,6 +21,7 @@ module To_test = struct
     p
   let wilcoxon_left_side x y =
     let (_, p, _) = M.wilcoxon ~side:M.LeftSide x y in
+    p
```
superbobry: This test wasn't compiling.

ryanrhymes: Thanks for fixing it 👍

ryanrhymes:
Many thanks for this PR, really great!

PR #49, opened 2017-06-13
Add Readme Chinese translation

jzstark:


ryanrhymes:
This is really good, many thanks 👍

Issue #47, opened 2017-06-08
feature request: rotate 3D plots


mars0i:
This might be planned anyway--I know that work on the plot routines are in progress--but it would be helpful to be able to specify the orientation of 3D plots.  Thanks for including the plot routines.  I just took a look at plplot, and I appreciate the way in which Owl greatly simplifies usage, as Tutorial 4 notes.

mars0i:
I think I see what additions would be needed.  I'm having trouble compiling Owl ("Unbound value Eigen.Dense.S.gemm" in owl_dense_common.ml--will investigate further), so for now I thought I'd just document what additions I think would be needed to allow different perspectives on 3D plots.

In owl_plot.ml, one would add the following.

In `type page {...}`:
```OCaml
mutable altitude : float;
mutable azimuth : float;
```
And somewhere in the same file:
```OCaml
let set_altitude h s = (h.pages.(h.current_page)).altitude <- s
let set_azimuth h s = (h.pages.(h.current_page)).azimuth <- s
```
and in the definition of `_prepare_page`, in the 3D plot section, in the call to `plw3d`, replace `33.` by `p.altitude`, and  `115.` by `p.azimuth`.

And add signatures `set_altitude` and `set_azimuth` to the .mli file.

mars0i:
Ah--looks like I forgot to install eigen.

ryanrhymes:
Yes, adding more fields to page type can be one solution. Another possible solution is incorporating the orientation parameters into the plot functions. You might already notice that Owl caches all the drawing operations then executes them one by one in the final stage. So it might be possible to `cache` that orientation parameters in the drawing function.

But this is just my current thought, I don't know myself whether it works or not. Or if that is really a cleaner solution comparing to add more fields to the page type.

mars0i:
OK, I'll look into those possibilities.

mars0i:
I have `altitude` and `orientation` working as fields in the page in the way I described above: https://github.com/mars0i/owl/blob/master/lib/owl_plot.ml .   I added functions `set_altitude` and `set_azimuth` to allow setting these values in the page structure.

The altitude and azimuth are used by the plplot function `plw3d`, which sets up some of the 3D parameters.  This is the only way to set the viewing perspective, as far as I know.  `plw3d` is already called in `_prepare_page` in owl_plot.ml, so I simply modified that call so that it would use the `altitude` and `azimuth` from the page.  

[...]

PR #52, opened 2017-06-23
owl_plot.ml,mli: add altitude, azimuth to page, 3D plot functions

mars0i:
Proposed solution to issue #47:

Added `altitude` and `azimuth` fields to the `page` type (`alt` and `az` here: http://plplot.sourceforge.net/docbook-manual/plplot-html-5.12.0/plw3d.html).  These  control the perspective from which `mesh` and `surf` plots are viewed.  

The two fields are initialized to default values in `_create_page` and are passed to `plw3d` in `_prepare_page`.

Also added functions `set_altitude` and `set_azimuth` to set the values using a `handle`, and added optional arguments to `mesh` and `surf` to specify altitude and azimuth.

### DIFF
--- a/lib/owl_plot.ml
+++ b/lib/owl_plot.ml
@@ -855,8 +855,8 @@ let surf ?(h=_default_handle) ?(contour=false) ?altitude ?azimuth x y z =
   (* prepare the closure *)
   let p = h.pages.(h.current_page) in
   let _ = p.is_3d <- true in
-  let _ = match altitude with Some a -> p.altitude <- a | None -> () in
-  let _ = match azimuth with Some a -> p.azimuth <- a | None -> () in
+  let _ = match altitude with None -> () | Some a -> p.altitude <- a in
+  let _ = match azimuth with None -> () | Some a -> p.azimuth <- a in
   let opt = match contour with
     | true  -> [ PL_FACETED; PL_MAG_COLOR; PL_BASE_CONT; PL_SURF_CONT ]
     | false -> [ PL_FACETED; PL_MAG_COLOR ]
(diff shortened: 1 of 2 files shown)


ryanrhymes:
This is a very good PR. Thank you very much @mars0i 

I did a couple of tests, the figures generated with this new feature looks cool! An example:

```ocaml
let x, y = Mat.meshgrid (-2.5) 2.5 (-2.5) 2.5 100 100 in
let z = Mat.(sin ((x * x) + (y * y))) in
let h = Plot.create ~m:1 ~n:3 "test_mesh.png" in
Plot.subplot h 0 0;
Plot.mesh ~h ~altitude:50. ~azimuth:120. x y z;
Plot.subplot h 0 1;
Plot.mesh ~h ~altitude:65. ~azimuth:120. x y z;
Plot.subplot h 0 2;
Plot.mesh ~h ~altitude:80. ~azimuth:120. x y z;
Plot.output h;;
```

The figure generated is as follows.

test_mesh

mars0i:
I'm very glad it seems good.  I'm very happy to contribute.  

I'm using Owl to run simulations that will play a crucial role in a conference presentation I'm giving next month, and the points I want to make will be much clearer with 3D plots.  The `altitude`/`azimuth` control allows me to shift the viewing perspective to one that will make the plots easier to understand.

Owl has made it very simple to create the plots I need.  Otherwise, I'd probably spend time reformatting data so that it could be used by the Lattice package in R.  That would be a lot more work.  (I'd never used plplot, but now that I've looked at it, I would rather avoid using it directly.)

Thank you!

Issue #57, opened 2017-07-04
Addtional 3D plot options


mars0i:
There are variations on 3D plots that Owl doesn't currently provide but that would be useful for me.  For example, this plot was created by using `[PL_DRAW_LINEY]` as the `opt` argument to `plmesh`:
[foo.pdf](https://github.com/ryanrhymes/owl/files/1122727/foo.pdf)
(The color variation in the plot is something I'll discuss in a different issue.)

I don't want to make Owl's plotting API overly complicated.  Its simplicity is what makes it so nice to use.  So I've thought about different ways to make `mesh` and `surf` more flexible.   This is my current thinking:

Ideally it's good to give users as much flexibility as possible with `mesh` and `surf`.  

(a) However, adding an optional argument for every option that can be passed (e.g. `PL_DRAW_LINEY`, etc. is confusing and would make the interface to `mesh` and `surf` a mess.  

(b) The number of optional arguments could be reduced by adding only one optional argument that allows users to pass something like `Plplot.([ PL_DRAW_LINEXY; PL_MAG_COLOR; PL_MESH ])`.  This is confusing, though, and interferes with the simplicity of the Owl interface.  

(c) Another option is to define new 3D plot options that use different plplot option lists.  However, it's also important to avoid writing many different plot functions with a lot of duplicate code.  (On the other hand, I do think that a little bit of duplicated code can be easier to understand rather than abstracting out too much of the code into common functions.)

So there's no ideal solution, other just than (d) not allowing more flexibility in 3D plotting.  

My inclination at this point is use option (b) to add an optional first argument to `mesh` and `surf`.  This argument allows passing a list of plplot options for `plmesh`, `plmeshc`, or `plsurf3d`.   It's not a nice interface, but then one can use this optional argument to define new 3D plotting functions as in (c).

For example, this is the definition of the function `plots2d_in_3d` that I used to generate the PDF linked above:
```OCaml
let plots2d_in_3d = mesh ~plplot_opt:Plplot.([PL_DRAW_LINEY])
```
(I couldn't think of a better name for this function.) The full revised code for `mesh` that I used (and `surf`, too) is here: https://github.com/mars0i/owl/blob/master/lib/owl_plot.ml#L877

What do you think?  I didn't want to submit a PR before seeing what your thoughts about this were.

ryanrhymes:
[...]

PR #64, opened 2017-07-09
Add option to spec for plplot 3D options plus code to use it

mars0i:
This is for issue #57.

I added a bit of text to the doc comments in the mli file.  Not sure if that's more than you want there.

The commit history is kind of odd looking (to me, at least).  Not worth explaining.

### DIFF
--- a/examples/owl_plot_examples.ml
+++ b/examples/owl_plot_examples.ml
@@ -1,0 +1,286 @@
+open Owl;;
+
+let f x = Maths.sin x /. x in
+let h = Plot.create "plot01.pdf" in
+Plot.set_foreground_color h 0 0 0;
+Plot.set_background_color h 255 255 255;
+Plot.set_title h "Function: f(x) = sine x / x";
+Plot.set_xlabel h "x-axis";
+Plot.set_ylabel h "y-axis";
+Plot.set_font_size h 8.;
+Plot.set_pen_size h 3.;
+Plot.plot_fun ~h f 1. 15.;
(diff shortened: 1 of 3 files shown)


mars0i:
I ran it through the examples in the tutorial. The only files that differed were the ones using random data, but visually they looked OK.  I'll PR the examples file, too.  I'm trying to separate different kinds of additions into different PRs.

ryanrhymes:
Great! Thanks @mars0i . It is an awesome PR and thanks for the contribution. Could you also revise the tutorial https://github.com/ryanrhymes/owl/wiki/Tutorial:-How-to-Plot-in-Owl%3F and contribute a small example there if possible. 👍 

Ideally, it is better not to expose raw Plplot configurations to Owl users in order to maintain a consistent plotting configuration. I will do some minor refactoring to wrap them up. but it is not a big issue at the moment. :)

mars0i:
Sure, I'd be happy to contribute an example.  I want to wait until the interface has stabilized. 

(I'm using Owl for a presentation at a week-long conference that starts a week from now.  The code has been changing pretty quickly--I'm not complaining, but I might need to stop contributing for a couple of weeks and just make use of my fork, which has some additional changes, until after the conference.  We'll see.  Working on Owl can be a nice way to take a break.)

ryanrhymes:
Sure, this is not urgent at all. I would also recommend using your own fork for the presentation. As you mentioned, the master branch is changing quite fast at the moment.

Good luck with your presentation.

PR #66, opened 2017-07-12
Fix email link and a couple of typos.

jasondavies:


ryanrhymes:
@jasondavies many thanks for fixing these, really appreciate it 👍

PR #69, opened 2017-07-16
update log plot and add image plot

jzstark:
Sorry for this late PR. Here is what it contains:

1. in the new version, `semilogx` and `semilogy` are only special cases of `loglog`, to which I  added an `axis` parameter.
2. Add a image plot. Please see here ([code](https://gist.githubusercontent.com/jzstark/9f49c77626533fb8e84a5131f67d4e05/raw/80da435d38312de63bb841a5af61853c88571775/examples.ml), [figure](https://gist.githubusercontent.com/jzstark/9f49c77626533fb8e84a5131f67d4e05/raw/80da435d38312de63bb841a5af61853c88571775/plot_026.png)) for an example. To remove the ticks/labels around the image, I add a `is_image` property to `page` in order for `_config_2d_axis` to return the proper code -1.

### DIFF
--- a/lib/owl_plot.ml
+++ b/lib/owl_plot.ml
@@ -1223,6 +1223,103 @@ let loglog ?(h=_default_handle) ?(spec=[]) ?x y =
   _add_legend_item p LINE line_style color marker color 0 color;
   if not h.holdon then output h
 
+  let semilogx ?(h=_default_handle) ?(spec=[]) ?x y =
+    (* TODO: express semilogx and semilogy with loglog *)
+    let open Plplot in
+    let y = Owl_dense_matrix.D.to_array y in
+    let n = Array.length y in
+    let x = match x with
+      | Some mtx -> Owl_dense_matrix.D.to_array mtx
+      | None     ->
+        Owl_dense_matrix.D.linspace 1. (float_of_int n) n
+        |> Owl_dense_matrix.D.to_array
+    in
+    let x = Array.map Owl_maths.log10 x in
+    _adjust_range h x X;
+    _adjust_range h y Y;
+    (* prepare the closure *)
+    let p = h.pages.(h.current_page) in
+    p.xlogscale <- true;
+    let color = _get_rgb spec p.fgcolor in
+    let r, g, b = color in
+    let marker = _get_marker spec "" in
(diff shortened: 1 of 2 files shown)


lib/owl_plot.ml
```diff
@@ -1560,6 +1578,38 @@ let scatterhist = None
...
+let image ?(h=_default_handle) mat=
+  let open Plplot in
+  (* compute necessary parameters *)
+  let width, height = Owl_dense_matrix.D.shape mat in
+  let num_col = Owl_dense_matrix.D.max mat in
+  let img = Owl_dense_matrix.D.to_arrays mat in
+  let width = float_of_int width in
+  let height = float_of_int height in
+  let num_col = int_of_float num_col in
+  (* specify the boundary of imageplot*)
```
ryanrhymes: use consistent format (* ... *)

ryanrhymes:
Thanks, this is really useful PR.  A couple of suggestions: 1) we perhaps need to change the function name `image` to something else if it only shows grayscale image; 2) `image` can be used to showcase the neural network examples in future; 3) I think `axis` can be part of the `spec`, so I will refactor the code a bit later, remember to `fetch upstream` to sync :)

Nice PR and great job! 👍

hcarty:
It may be worth using `plimagefr` for this as it would allow more control over the color range used.

jzstark:
@hcarty  Thanks for your suggestion. Currently I'm trying to use `plplot` to draw color images. I find functions that can manipulate/load colormap palette and map color index to continuous colors, but not a function that can map RGB color back to color index, since my understanding is that `plimage` and `plimagefr` only accept one index color matrices as input, while the image data I have consist of 3 matrices, one for each color channel. I've checked related documents but must have missed something. It will be really appreciated if any suggestion about it could be kindly provided. Thanks!

hcarty:
@jzstark plplot does not have a function to map a specific color to a specific matrix element.  `plimagefr` gives more control over the color range used within a continuous color palette but it doesn't allow directly mapping colors.

It's something that would be good to have in plplot in general so if you want to get your hands dirty in some C then I suspect that a contribution would be welcome!  I haven't been actively involved in the project for a while though unfortunately so I can't offer much direct insight.

PR #71, opened 2017-07-17
fix image plot bug

jzstark:
Thank for pointing out this problem. As you may have found out, the input matrix need to be transposed to be positioned correctly. I will also update/rotate the example file `durer.txt` (which is actually now a 600 * 900 matrix) accordingly.

### DIFF
--- a/lib/owl_plot.mli
+++ b/lib/owl_plot.mli
@@ -310,5 +310,5 @@ val qqplot : ?h:handle -> ?spec:spec list -> ?pd:(float -> float) -> ?x:dsmat ->
 
 val image : ?h:handle -> dsmat -> unit
 (**
-  [image x] display a m * n matrix [x] xas image. Each element in the matrix is of range 0 ~ 255.
+  [image x] display a [m x n] matrix [x] as image. Each element in the matrix is of range 0 ~ 255.
  *)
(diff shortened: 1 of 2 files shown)


lib/owl_plot.mli
```diff
@@ -310,5 +310,5 @@ val qqplot : ?h:handle -> ?spec:spec list -> ?pd:(float -> float) -> ?x:dsmat ->
 
 val image : ?h:handle -> dsmat -> unit
 (**
-  [image mat] display a m * n matrix as image. Each element in the matrix is of range 0 ~ 255.
+  [image x] display a m * n matrix [x] xas image. Each element in the matrix is of range 0 ~ 255.
```
ryanrhymes: m * n -> [m x n]
xas image -> as an image

ryanrhymes:
@jzstark I am not sure your new implementation is correct either. Try the following code:

```ocaml
let z0 = Mat.create 100 100 200. in
let z1 = Mat.zeros 100 100 in
let z = Mat.((z1 @|| z0) @= (z1 @|| z1)) in
Plot.image z;;
```
The right square should be top right. As I said before, it is flipped! So transpose will not solve the problem.

Please make sure your code well tested before PR, thanks.

PR #73, opened 2017-07-17
fix image-flip bug of imageplot

jzstark:
So sorry for the untested PR.  Here is how I propose to fix this problem and why:

If we `image` the input matrix as is, it will recognise the top left corner of the matrix as origin, with x-axis along rows and y-axis along columns, which is exactly how a matrix locate its elements. So I rotate the input matrix 90 degree clockwise (I failed to find a `Mat.rotate` or similar function from the doc so just implement a simple one) and then plot it as image.   I test with the code:

```ocaml
let z0 = Mat.create 100 100 200. in
let z1 = Mat.zeros 100 100 in
let z = Mat.((z1 @|| z0) @= (z1 @|| z1) @= (z1 @|| z1)) in
Plot.image z
```

The white rectangle (though not a square as we expected  --- I'll look into this problem later) is correctly shown on the upper left corner.

### DIFF
--- a/lib/owl_plot.ml
+++ b/lib/owl_plot.ml
@@ -1593,14 +1593,24 @@ let scatterhist = None
 
 (* other plots *)
 
+let _rotate270_ x =
+  let w, h = Owl_dense_matrix.D.shape x in
+  let mat = Owl_dense_matrix.D.zeros h w in
+  for i = 0 to w - 1 do
+    for j = 0 to h - 1 do
+      Owl_dense_matrix.D.set mat j (w - 1 - i) (Owl_dense_matrix.D.get x i j)
+    done
+  done;
+  mat
+
 let image ?(h=_default_handle) x =
   let open Plplot in
+  (* rotate the matrix 90 degree clockwise to be shown correctly as image *)
+  let x = _rotate270_ x in
   (* compute necessary parameters *)
-  let height, width = Owl_dense_matrix.D.shape x in
+  let width, height = Owl_dense_matrix.D.shape x in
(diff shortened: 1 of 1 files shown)


lib/owl_plot.ml
```diff
@@ -1593,14 +1593,24 @@ let scatterhist = None
 
 (* other plots *)
 
+let _rotate270_ x =
+  let w, h = Owl_dense_matrix.D.shape x in
```
ryanrhymes: I will implement a `flip` function in Matrix module later today. Before that, a better way is to use `slicing` function. E.g.,

```ocaml
Mat.slice [[-1;0];[-1;0]] x
```

Less code and way faster :)

jzstark: Thanks! Using `slice` rotate the matrix for 180 degree quite concisely. I'm looking at how a -90 degree rotation can be achieved with it and other existing matrix manipulation functions

ryanrhymes: OK, I see. Then probably you need `transpose` + `slice`. I think the better solution is I implement `rotate` function in the core. I will merge this PR then replace your "_rotate_" function later.

ryanrhymes:
Good PR, looking forward to the coloured version 👍

lib/owl_plot.ml
jzstark: Thanks a lot. The `transpose` + `slice` can do exactly the rotate magic required:

```ocaml 
let _rotate270_ x =
  x |> Mat.transpose |> Mat.slice [[]; [-1;0]] 
```
But of course a `rotate` function would be great to have :)

PR #76, opened 2017-07-20
Matrix kron

jotterbach:
This adds some functionality to compute the Kronecker product between to matrices of type float. @ryanrhymes I have some troubles making this fully generic. Maybe you can offer some advice?


ryanrhymes:
@jotterbach many thanks for the PR, great job. Using `Owl_dense_common._mul_elt` instead of `*.` makes the function generic, so I only changed about one line of your code :)

PR #78, opened 2017-07-25
Fixing `l2norm` for complex arrays

jotterbach:
When calculating the squared L2 norm of complex vectors the result can be negative giving rise to `NaN` when taking the `sqrt`. L2 norms are positive-semidefinite at all times. This fix adjusts the accumulator function for the L2 norm of complex numbers.

* Adjust accumulator for complex numbers
* Add unit test for complex vector

### DIFF
--- a/lib/owl_dense_common_c.c
+++ b/lib/owl_dense_common_c.c
@@ -884,14 +884,14 @@ value cp_two_doubles(double d0, double d1)
 #define FUN5 complex_float_l2norm_sqr
 #define INIT float r = 0.
 #define NUMBER complex_float
-#define ACCFN(A,X) (A += X.r * X.r - X.i * X.i + 2 * X.r * X.i)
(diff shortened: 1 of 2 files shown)


jotterbach:
@ryanrhymes I was looking at the stack trace of the failing Travis build and it seems to happen in a file that this PR does not touch. Also I can't seem to reproduce this failure locally. Can you advise?

ryanrhymes:
@jotterbach You are absolutely right, many thanks for fixing it! 

I cannot reproduce the Travis failure either, it doesn't seem like caused by the committed code.

ryanrhymes:
The Travis failure was due to the recent API changes in Alcotest. More precisely, due to this: https://github.com/mirage/alcotest/commit/eac111f2a09c632994007180c61ee4480ab92d36

Thanks for the great PR again 👍

PR #81, opened 2017-08-01
Update MNIST and LSTM examples

jzstark:
- Update `test_lstm.ml` and `test_neural.ml` in examples with new API (e.g. lstm now takes one rather than two parameters) and new functionality (e.g. model save/load and mnist dataset visualisation).  
- One small problem here is that in my `utop` I actually cannot load the `Owl_nlp_utils` module.
- Two other experimental model scripts are removed in this PR for they are yet incomplete as examples.


ryanrhymes:
Thanks Jianxin, cool PR 👍  Since the Zoo system is up and running, we might start using it for the examples, scripting is much lighter and does not require compilation.

ryanrhymes:
I have changed all the examples to the new zoo system. All seem working on my machine, maybe fetch the upstream to sync your branch and give it a try?

jzstark:
@ryanrhymes I've just tried and all examples work fine, except for two hard-coded dataset paths in `test_lda.ml` and `test_tfidf.ml`. I've fixed them in a new push. Besides, it seems that I cannot download the `news.*` data through `Dataset` as required in these two examples. 

btw, I found that the (stupid) reason I cannot get the three `owl_*.sh` scripts is that I didn't use `sudo` when `make`.

ryanrhymes:
> @ryanrhymes I've just tried and all examples work fine, except for two hard-coded dataset paths in `test_lda.ml` and `test_tfidf.ml`. I've fixed them in a new push. Besides, it seems that I cannot download the `news.*` data through `Dataset` as required in these two examples. 

Great, I hope the new zoo system will make prototyping and experiments easier in future. The news dataset was used in Kvasir project, I will add it to Owl's dataset repo later this week after some cleaning up.

> btw, I found that the (stupid) reason I cannot get the three `owl_*.sh` scripts is that I didn't use `sudo` when `make`. 

I see. Yes you need admin privilege in order to write bin folder.

PR #83, opened 2017-08-02
Fixed module paths to Gsl.Fft

mmottl:
The latest GSL release packages the library differently (using `jbuilder`), making it incompatible with Owl.  `Gsl_fft` was never supposed to be accessible, but leaked through anyway.  The correct way is to always use `Gsl.?` for accessing parts of the packed library.


ryanrhymes:
Great! Markus, many thanks for fixing this :)

mmottl:
There is also another pull request in the OPAM repository for fixing version constraints: https://github.com/ocaml/opam-repository/pull/9994

It's not compiling right now, apparently due to some download problems with external libraries.  If it continues to have problems, you may want to check it to see whether there is a deeper problem with Owl dependencies.

PR #87, opened 2017-08-18
Update Dataset and documents

jzstark:
Here is a summary of this RP:

- update CIFAR10 dataset in `Dataset` module
- update a draft documents for `owl_neural_graph` module
- try using zoo script in the MNIST example 
- fix some other issues: add new dependency in Readme file, fix some typos, and update existing plot module doc, etc.


ryanrhymes:
Great PR, any improvements on the documentation is highly appreciate :)

ryanrhymes:
a side note, try to keep the example code concise and remove the code that have been commented out. another thing i noticed is that the mnist_cnn example converges much slower than my original version, probably better to compare my original one : https://gist.github.com/ryanrhymes/642a756a4c10609fd34ed9b60290b5c9

When updating the example, make sure it will work better not worse.

PR #97, opened 2017-09-01
fix Mat.std function

jzstark:
A summary of this PR: 
- Fix one bug in the `std` function of matrix: the denominator should be sqrt-ed before being used
- A minor update of `CONTRIBUTING.md`


ryanrhymes:
Thanks for fixing this, much appreciated!

Issue #101, opened 2017-09-07
Simpler instructions for building master


dbuenzli:
The instructions [here](https://github.com/ryanrhymes/owl#installation) seem needlessly long winded. You can simply instruct users to: 

```
opam depext owl
opam pin add owl --dev-repo
```

mseri:
This does not seem to install the `owl` binary and script. But it should not hard to fix

mseri:
Also depext is not installing some required dependencies (`liblapacke-dev` and `libopenblas-dev`)

dbuenzli:
> Also depext is not installing some required dependencies (liblapacke-dev and libopenblas-dev on ubuntu)

The the constraints in the opam file of the repo should be updated.

mseri:
Yes. I want to add them for ubuntu, debian and fedora based distribution, I need to prepare some docker environments to check which dependencies are needed

ryanrhymes:
Owl has changed to jbuilder, the building and installation seems fine now on my machine. In case you spot any issues, please let me know or help me to solve it, really appreciate it.

There might be some glitches but all will be solved soon. The documentation will be also be updated accordingly in near future.

PR #102, opened 2017-09-08
Link the math library when compiling Owl and install the opam binaries to the correct place

mseri:
This should fix the `undefined symbol` error on Linux machines (https://github.com/ryanrhymes/owl/issues/61).

The second part of the PR makes sure that the binaries are installed in the opam binary folder, fixing an issue I have encountered both in ubuntu and debian.

It would be nice to use the ocaml-docker image to enable tests in different linux distribution (and add the correct depext section to the opam file)


ryanrhymes:
@mseri Many thanks for fixing this, really appreciate. I have been trying out jbuilder this weekend, if it goes well, I might change to jbuilder in the future. The compilation seems faster.

mseri:
I was playing with it as well. Also splitting the top level and rewriting the pretty printers. Let me know if you decide to port to jbuilder and need help

ryanrhymes:
That's great. I already ported most of owl to jbuilder, I spent quite some time in sorting out the installation issue due to https://github.com/ocaml/opam/commit/718d6198e338069852dcdde067f56df236235370#diff-5508a1f41811a003b062a9d606062507 Thanks David @dra27 for the help. Currently, we need to use `opam.2.0.0~beta4` for installation. I have also split out the toplevel as Daniel @dbuenzli suggested.

I still need to fix the Zoo system a bit. But I think I can finish a working system with jbuilder by tomorrow. If you can help in pretty-printing, that would be really awesome. At the moment, the pretty printing still uses Markus Mottl code (with a bit minor tweak), I think it needs quite some redesign at some point (especially after I extend owl to incorporate integer types).

dbuenzli:
> Currently, we need to use opam.2.0.0~beta4 for installation.

Are you sure `jbuilder` uses `opam-installer` tool if the build is opam-managed ? I think you can still use previous opam versions if that's not the case, they should interpret the `.install` file fine. This is yet another reason why the workflow for installing master should be that of #101 rather than entice people to install owl in their local prefix.

dra27:
@dbuenzli is right - as long as your opam file doesn't explicitly run `jbuilder install` you should be fine. The problem with opam-installer only happens if users are installing owl manually.

ryanrhymes:
@dbuenzli @dra27 I see, the https://github.com/ryanrhymes/owl/tree/jbuilder seems working fine now. There is another thing still puzzles me a bit: 

if there are multiple libraries in one repo, jbuilder needs one .opam file for each. Do I keep .opam in the corresponding subdirectories, or do I keep them all in the root? Jbuilder seems to complain if I maintain them all in the root, I still haven't figured out how https://github.com/ocsigen/js_of_ocaml manages to do that.

For other libraries, it is said they need to be prefixed ... should I name it like `owl-zoo.opam` or `owl_zoo.opam`, or does it matter at all?

mseri:
They can be all in the root, how does it complain? You probably need to add (package owl) or (package owl_zoo) to the executables.

I think ’owl-zoo.opam’ is more standard but either should be fine

ryanrhymes:
@mseri awesome, I finally understood the complaints `Error: I can't determine automatically which package this (install ...) ...`

Yes, it is now solved as you have suggested, great 👍

PR #105, opened 2017-09-14
Change dataset to ndarray format

jzstark:
Change format of the datasets that are in `matrix` format to `ndarray`.


ryanrhymes:
These changes are not necessary because Matrix and Ndarray are the same type.

Issue #107, opened 2017-09-23
plots: more flexible area/fill function?


mars0i:
I want to fill a region between two curves--i.e. all points above the lower curve and below the upper curve should be filled.  `area` is not designed for this, I believe; it adds endpoints at y=0 so that `plfill` will close the curve along the x axis.  I don't think Owl has a function that does what I want, but I may be wrong.

Does a "fill between curves function" seem likely to be useful to others?  I could see writing it like `area`, but without the additional y=0 points.  Then it would be the user's responsibility to concatenate their vectors in the right way for `plfill`.

 Or the function could accept, for example, matrices whose rows or columns would be automatically concatenated to form the curves to be filled.  There are variations on ths scheme--passing lists of vectors, for example.  

There may be tricky parts of the problem that I have not thought through yet.

ryanrhymes:
You are right, I don't remember there is such a function in Owl at the moment to allow you to do that. There are several solutions:

- Use `area` to draw one function with colour different from background, then draw another function with the same colour of the back to mask it. You can use `min2` and `max2` to figure out the upper bound and lower bound. 

- Maybe it is about time to introduce transparency `Alpha` into `spec` list? This may also be useful if two functions intersect with each other.

I agree this is a useful function. I think it is worth adding a general solution for this. I will have a look at how other software to do that. But in case you have any material to share or any suggestions, always welcome. 

Maybe point 1 can be extended to a general solution?

mars0i:
Thanks Liang.  I thought about using the background color to mask the unwanted fill.  That would work but seems inelegant as a long-term solution, since `plfill` already has the capability to fill the region between two curves.  I agree that `Alpha` seems like it could be very useful, although I don't think I need it yet fwiw.

`region` below is one attempt at a solution.  It's almost the same as `area`.  All of the changes are in the first several lines.  The additional trick is that if the function is passed matrices with two rows rather than one row, it treats one row as the upper curve, and the other as the lower curve (it doesn't matter which is which).  The function concatenates the two rows, reversing the second one so that `plfill` can treat the extended vectors as defining a polygon to fill.  Or if the matrices passed have only one row, the function will behave like `area`.  (I moved the code that adds the endpoints at *y=0* into a separate function, but it could stay in the main function.)
```ocaml
let _circular_concat x =
  Owl_dense_matrix.D.(concat_horizontal (row x 0) (reverse (row x 1)))

let _add_ends_at_zero x y =
[...]

PR #111, opened 2017-09-26
added draw_polygon

mars0i:
Identical to `area` but with four lines deleted, and a new doc comment.

I think I'm going to try also doing a second PR that's the same but that pulls out the code that `area` and `draw_polygon` into an internal function.  If github will allow two separate PRs (??), then you can choose.

(I understand the value of having different functions that duplicate code.  It's simpler to understand, and if you want to change one but not the other, it's easier.  However, in this case `draw_polygon` really is `area` with four lines deleted, so maybe it's reasonable to have them call the same internal function to run the code that's the same in both functions.)

This is in response to #107.

### DIFF
--- a/src/owl/owl_plot.ml
+++ b/src/owl/owl_plot.ml
@@ -1011,75 +1011,51 @@ let bar ?(h=_default_handle) ?(spec=[]) y =
   if not h.holdon then output h
 
 
-let area ?(h=_default_handle) ?(spec=[]) x y=
+let _area_fill h spec x y =
   let open Plplot in
+  _adjust_range h x X;
+  _adjust_range h y Y;
+  (* prepare the closure *)
+  let p = h.pages.(h.current_page) in
+  let color = _get_rgb spec p.fgcolor in
+  let r, g, b = color in
+  let line_style = _get_line_style spec 1 in
+  let fill_pattern = _get_fill_pattern spec 0 in
+  (* drawing function *)
+  let f = (fun () ->
+    let r', g', b' = plgcol0 1 in
+    plscol0 1 r g b;
+    plcol0 1;
+    pllsty line_style;
+    plline x y;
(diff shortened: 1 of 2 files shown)


mars0i:
OK, as I suspected, I can only have one PR live at the same time.  If you want, take a look at this:
https://github.com/mars0i/owl/blob/master/src/owl/owl_plot.ml#L1014
It pulls out the code that's common to `area` and `draw_polygon` and puts it in a function `_area_fill` that `area` and `draw_polygon` call.  I'll update this PR with that code if you prefer that coding style for `area` and `draw_polygon`.  Either way of writing these functions seems fine to me.

ryanrhymes:
Thanks Marshall. IMO, the [first PR](https://github.com/ryanrhymes/owl/pull/111/commits/8cf519c241d1b42461abcc6e714d613efbc4b6a1) looks better. I know there is some overlap between the two functions but it looks clearer and each is self-contained. Let's abstract the common part out in future if there are more similar functions added.

ryanrhymes:
@mars0i Can you revert to the first PR so I can merge?

mars0i:
Oh, sorry.  I forgot that when I pushed the change it would go into the PR.  That behavior goes against my intuitions that I continually forget, even when I have just been reminded of it.

ryanrhymes:
Many thanks @mars0i , this is a really nice PR, great job. 👍 

I am still open to discussion about introducing the specific functions you mentioned. We just need a bit time to have a look at how others do things. But this `draw_polygon` is definitely very general and much needed imo.

mars0i:
That all sounds good.  Thanks.  I'm very happy to have this in the master branch, and will be using it immediately.  It's the sort of thing that's difficult to add without inserting it into owl_plot.ml, where all of the necessary definitions have been set up, but maintaining a separate branch eventually becomes problematic.

Issue #113, opened 2017-09-29
Strange interaction between plot functions


mars0i:
If I combine `plot` with `area`, the plotted area is incorrect if the `plot` line is on or near the displayed region, and `plot` is called before `area`.  I first noticed this kind of behavior with `draw_polygon` instead of `area`, but I did more testing with `area` since using it is a little bit simpler--there is only one defined edge.  I haven't experimented with other combinations involving `area` or `draw_polygon`.

Experiment 1:
```ocaml
module M = Owl.Mat;;
module P = Owl.Plot;;

let xs = M.sequential 1 10;;
let ys = M.map (fun x -> x**2.)  xs;;

(* An area plot: *)
let h = P.create "foo.pdf" in P.(area ~h xs ys; output h);;

(* Let's add line along the top edge: *)
let h = P.create "foo.pdf" in
P.(area ~h xs ys;
   plot ~h ~spec:[RGB(0,0,255); LineWidth 5.] xs ys;
   output h);;

(* Try the same thing but call plot for the line first: *)
let h = P.create "foo.pdf" in
P.(plot ~h ~spec:[RGB(0,0,255); LineWidth 5.] xs ys;
   area ~h xs ys;
   output h);;
(* Where did the blue line go? *)

(* Let's make the line wide.  First call area and then call plot. *)
let h = P.create "foo.pdf" in
P.(area ~h xs ys;
   plot ~h ~spec:[RGB(0,0,255); LineWidth 50.] xs ys;
   output h);;
(* It looks as if the line is centered along the edge of the area;
 * that makes sense. *)

(* Now lets try calling plot first: *)
let h = P.create "foo.pdf" in
P.(plot ~h ~spec:[RGB(0,0,255); LineWidth 50.] xs ys;
   area ~h xs ys;
   output h);;
(* Very strange.  It looks as if the plot line is missing, and the entire `area` region
 * has an extra wide edge on all sides, with some diagonal corners.  *)
```

Experiment 2:  Same as experiment 1, but for `plot`, give arguments `xs xs` rather than `xs ys`, so that a straight line is (supposed to be) drawn over the area.  With `plot` second, it behaves as expected.  With `plot` first, and `LineWidth 50.`, the image seems to be identical to the last example in experiment 1.

Experiment 3:
```ocaml
let ys' = M.map (fun x -> (x**2.) +. 10.)  xs;;
```
[...]

PR #114, opened 2017-09-30
Fix bug in plot, loglog: let within if was capturing rest of closure

mars0i:
plot, loglog: added parentheses around let in 'if marker'

In the closure definition in `plot`, because the `let` in scope of `if marker ...` wasn't wrapped in parentheses, the `let` extended to the rest of the closure, and nothing else got executed if the `if` test failed. It's also clear from the indentation that this isn't what was intended.  Plot settings weren't getting rrestored to their previous values at the end of the closure, which caused some of the settings to carry over to the next closure when it was executed.

This fixes issue #113.  I don't know why the PR includes some old commits from my previous PR.  Let me know if I need to redo this one.

It seems clear that the same problem exists in loglog, so I fixed that. (EDIT: At first I couldn't run `loglog` just to test that I hadn't broken something, but now I figured out how to do that.)

(I haven't noticed other functions in owl_plot.ml that have this problem.  I searched through all 'if' expressions that use 'marker'.  Difficult to search for an if followed by a let on the next line.)

(This is the `loglog` error:
```ocaml
# let xs = M.sequential 1 10;;
val xs : M.mat =
   C0 C1 C2 C3 C4 C5 C6 C7 C8 C9
R0  0  1  2  3  4  5  6  7  8  9

# let h = Owl.Plot.create "foo.pdf" in Owl.Plot.(loglog ~h xs ; output h);;
Exception: Gsl__Error.Gsl_exn (Gsl__Error.EDOM, "domain error").
Raised by primitive operation at unknown location
Called from file "src/owl/owl_maths.ml", line 89, characters 14-21
Called from file "array.ml", line 91, characters 21-40
Called from file "src/owl/owl_plot.ml", line 1236, characters 41-68
Called from file "//toplevel//", line 1, characters 47-59
Called from file "toplevel/toploop.ml", line 180, characters 17-56
```
I'll try to figure out whether I'm doing something wrong and file an issue if need be.)

### DIFF
--- a/src/owl/owl_plot.ml
+++ b/src/owl/owl_plot.ml
@@ -640,9 +640,10 @@ let plot ?(h=_default_handle) ?(spec=[]) x y =
     plschr marker_size 1.;
     if line_style > 0 && line_style < 9 then
       pllsty line_style; plline x y;
-    if marker <> "" then
+    if marker <> "" then (
       let x', y' = _thinning x, _thinning y in
-      plstring x' y' marker;
+      plstring x' y' marker
+    );
     (* restore original settings *)
     plschr c' 1.;
     plwidth old_pensize;
@@ -1259,9 +1260,10 @@ let loglog ?(h=_default_handle) ?(spec=[]) ?x y =
     plschr marker_size 1.;
     if line_style > 0 && line_style < 9 then
       pllsty line_style; plline x y;
-    if marker <> "" then
+    if marker <> "" then (
       let x', y' = _thinning x, _thinning y in
-      plstring x' y' marker;
+      plstring x' y' marker
+    );
     (* restore original settings *)
     plschr c' 1.;
     plwidth old_pensize;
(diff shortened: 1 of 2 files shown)


jzstark:
@mars0i Thank you so much for finding out and fixing this problem! I wrote the `loglog` function based on `plot` and I think that's why they share the same problem.

ryanrhymes:
This is definitely a bug, many thanks for fixing it @mars0i 

For the second issue, I think the reason is the `0` in `xs`, which `loglog` will call `Maths.log 0` which will throw that exception.

mars0i:
Ah, I see.  That makes sense.

ryanrhymes:
you can instead try with `let x = Mat.sequential ~a:1. 1 10;;`

mars0i:
Thanks.  I must have noticed the optional arguments to `sequential` at some point, but completely forgot about them.

Issue #117, opened 2017-10-16
row function allows modification of argument but col doesn't


mars0i:
Modifying the result of `row` affects its argument matrix, while modifying the result of `col` does not. My guess is that this difference has to do with efficiency and the underlying ndarray structure.  It seems confusing that `row` and `col` behave differently.  I don't know whether it would make sense to modify one of these functions.  One option is to leave the definitions as they are now but change the doc comments so that users know about the difference.
```ocaml
# let m = Mat.sequential 2 2;;
val m : Mat.mat =
  C0 C1
R0  0  1
R1  2  3

# let r0 = Mat.row m 0;;
val r0 : Mat.mat =
  C0 C1
R0  0  1

# let c0 = Mat.col m 0;;
val c0 : Mat.mat =
  C0
R0  0
R1  2

# Mat.set r0 0 1 42.;;
- : unit = ()
# Mat.set c0 1 0 42.;;
- : unit = ()
# m;;
- : Mat.mat =
  C0 C1
R0  0 42
R1  2  3
```

ryanrhymes:
@mars0i thanks for pointing it out, this actually relates to our previous discussion https://github.com/ryanrhymes/owl/issues/98

`row` returns a "pointer" to the same memory without copy whereas `col` returns a copy. You are right this is due to performance consideration. Because Owl sticks to c memory layout so taking a row is taking continuous memory block, much faster than retrieving a `col`.

I think the best option for now is emphasising that in the documentation.

mars0i:
Ah, sorry, I had forgotten about that part of the conversation.  PR with modified comments coming in a moment.

nilsbecker:
i think adding in the docs is good but i would still vote for also changing the names to prevent nasty bugs. afaik, julia has a convention of adding a ! to functions which mutate; maybe some convention for functions which return a view would be good for owl? 

i'm a bit sensitive to hidden mutability since i recently found a bug resulting from using the lacaml `gesv` function without paying attention to mutation of one of the arguments. after a month or so. even though i was aware of the general possibility. this made me think that in a language like ocaml where pervasive mutability is usually avoided, shared mutable values should be flagged whenever possible.

but maybe a more comprehensive view interface is planned? that would be a better time to change the api then.

mars0i:
[...]

PR #120, opened 2017-10-18
row, col doc comments: added notes about view vs. copy

mars0i:
I thought it would be good for the `row` doc comment to note the difference with `col`, and vice versa, since it's natural to assume that they are the same.  Let me know if this is too much text.  For #117.


ryanrhymes:
Marshall, thank you for clarifying the difference between `row` and `col` 👍

PR #121, opened 2017-10-20
In-place mutation: Allow limited broadcast.

bluddy:
In X += Y, Y can be broadcasted to X so long as it's strictly smaller. X
cannot broadcast.


ryanrhymes:
Yotam, thanks for the PR. I think a better solution is to check the shape of two operands to decide whether broadcast can be performed, rather than introducing another quite similar broadcasting function. I will temporarily close this PR since that module is undertaking quite drastic changes at the moment.

PR #126, opened 2017-10-28
Minor fixes to doc comments of file save/load functions

mars0i:
Fixed typo in `load` doc comment by deleting "sparse".  

While I was in that part of the file, I also added "tab-delimited" to the doc comments for `save_txt` and `load_txt`.  That seemed useful so that the user doesn't have to write a file to find out what format it will be in.  Let me know if it's better to leave "tab-delimited" out.


ryanrhymes:
Marshall thanks for improving the documentation, highly appreciate it.

PR #130, opened 2017-11-19
Fix build with OCaml 4.06.0 (and -safe-string)

djs55:
OCaml 4.06.0 uses `-safe-string` by default and so we have to distinguish between immutable `string` and mutable `bytes`.

This PR fixes the build for 4.06.0 and switches on `-safe-string` for all OCaml versions.

### DIFF
--- a/bin/jbuild
+++ b/bin/jbuild
@@ -8,6 +8,7 @@
     owl
     owl_zoo
   ))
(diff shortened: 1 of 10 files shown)


ryanrhymes:
Cool, great PR. David, many thanks for fixing this, really appreciate it! 💯

PR #133, opened 2017-11-25
Fix opam metadata

rgrinberg:
* add jbuilder as a dependency
* remove package names from opam files
* add build-test command


ryanrhymes:
@rgrinberg Cool, this is really useful PR! 💯

PR #136, opened 2017-12-10
Add Kolmogorov-Smirnov Tests to the Stats Module

jsthomas:
This change adds one and two sample Kolmogorov-Smirnov tests to `Owl.Stats`. To create my tests, I relied on a comparison with SciPy (along with reading papers/books). As a fair warning, I'm still pretty new to OCaml and Owl; apologies in advance if there are some style/mechanics issues in this PR.


mars0i:
It's very nice to have K-S tests--thanks.  I don't need them often, and don't have a need that I foresee at the moment, but I have used them in the past.

ryanrhymes:
Awesome PR 💯  I have been wanting to include K-S test into `Stats` for a very long time.  The code looks fine and the test and doc are very nice. Thank you Joe. 👍

PR #140, opened 2017-12-19
Added mpow function for iterating dot product of matrix with itself

mars0i:
Exponent is a float, but currently only handles whole-number floats as exponents.  An exception is raised otherwise.

Questions:

- Should the ocamldoc mention that only whole numbers are allowed?
- Should `mpow` handle the exponent 0.0, returning an identity matrix of the correct size?
- Have I modified all and only the right files?  I followed existing patterns, adding the signature to a number of files.
- Is the long, mostly irrelevant commit history annoying?  All of the actual remaining changes only concern `mpow`.  I can try to clean it up.  (gasche showed me some ways to do that with some Batteries PR's I submitted.)


mars0i:
Oops--didn't mean to commit a new `jbuild`.  I'll fix that.

mars0i:
Closing this PR.  Will submit one that's not so messy, if I can.

PR #141, opened 2017-12-20
add mpow function: iterate dot prod of matrix with itself

mars0i:
(I took the extreme step of deleting my fork and making a new one.  Trying to clean up a git history is a pita.  This seemed easier.  I hope it causes no problems.)

This PR adds this function:
```
(** [mpow x r] returns the dot product of square matrix [x] with itself [r] times. *)
```

The exponent `r` is a `float`, but must be a nonzero natural number.  An exception is raised otherwise.

Questions:

- Should the ocamldoc mention that only whole numbers are allowed?
- Should `mpow` handle the exponent 0.0, returning an identity matrix of the correct size?
- Have I modified all and only the right files?  I followed existing patterns, adding the signature to a number of files in order follow the same patterns (and to cause `mpow` to be included `Owl.Mat`).  I wasn't sure whether what I did in files other than src/owl/owl_dense_ndarray_generic.ml*` was correct.

### DIFF
--- a/src/owl/jbuild
+++ b/src/owl/jbuild
@@ -22,6 +22,7 @@
   ))
   (c_library_flags (
     -L/usr/local/lib/gcc/7
+    -L/usr/local/opt/openblas/lib
     -lopenblas
     -lgfortran
     -lm
--- a/src/owl/jbuild
+++ b/src/owl/jbuild
@@ -22,6 +22,7 @@
   ))
   (c_library_flags (
     -L/usr/local/lib/gcc/7
+    -L/usr/local/opt/openblas/lib 
     -lopenblas
     -lgfortran
     -lm
--- a/src/owl/jbuild
+++ b/src/owl/jbuild
@@ -22,7 +22,6 @@
   ))
   (c_library_flags (
     -L/usr/local/lib/gcc/7
-    -L/usr/local/opt/openblas/lib
     -lopenblas
     -lgfortran
     -lm
--- a/src/owl/owl_dense_matrix_c.mli
+++ b/src/owl/owl_dense_matrix_c.mli
@@ -589,6 +589,8 @@ val scalar_pow : elt -> mat -> mat
 
 val pow_scalar : mat -> elt -> mat
 
+val mpow : mat -> float -> mat
+
 val cov : ?b:mat -> a:mat -> mat
 
 val kron : mat -> mat -> mat
--- a/src/owl/owl_dense_matrix_d.mli
+++ b/src/owl/owl_dense_matrix_d.mli
@@ -592,6 +592,8 @@ val scalar_pow : elt -> mat -> mat
 
 val pow_scalar : mat -> elt -> mat
 
+val mpow : mat -> float -> mat
+
 val atan2 : mat -> mat -> mat
 
 val scalar_atan2 : elt -> mat -> mat
--- a/src/owl/owl_dense_matrix_generic.mli
+++ b/src/owl/owl_dense_matrix_generic.mli
@@ -1401,7 +1401,8 @@ val pow_scalar : ('a, 'b) t -> 'a -> ('a, 'b) t
 (** [pow_scalar x a] *)
 
 val mpow : ('a, 'b) t -> float -> ('a, 'b) t
-(** [mpow x r] returns the dot product of [x] with itself [r] times. *)
+(** [mpow x r] returns the dot product of square matrix [x] with 
+  itself [r] times. *)
 
 val atan2 : (float, 'a) t -> (float, 'a) t -> (float, 'a) t
 (** [atan2 x y] computes [atan2(a, b)] of all the elements in [x] and [y]
--- a/src/owl/owl_dense_matrix_generic.mli
+++ b/src/owl/owl_dense_matrix_generic.mli
@@ -1402,7 +1402,10 @@ val pow_scalar : ('a, 'b) t -> 'a -> ('a, 'b) t
 
 val mpow : ('a, 'b) t -> float -> ('a, 'b) t
 (** [mpow x r] returns the dot product of square matrix [x] with 
-  itself [r] times. *)
+  itself [r] times, and more generally raises the matrix to the
+  [r]th power.  [r] is a float that must be equal to an integer;
+  it can be be negative, zero, or positive. Non-integer exponents
+  are not yet implemented. *)
 
 val atan2 : (float, 'a) t -> (float, 'a) t -> (float, 'a) t
 (** [atan2 x y] computes [atan2(a, b)] of all the elements in [x] and [y]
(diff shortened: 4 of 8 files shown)


mars0i:
Urgh!  How did jbuild get back into the commits?

ryanrhymes:
Great , many thanks!

- Should the ocamldoc mention that only whole numbers are allowed?
  * Yes, it is better to mention that fraction number is not implemented yet.
- Should `mpow` handle the exponent 0.0, returning an identity matrix of the correct size?
  * Yes, it should return an identity matrix, right? How about negative numbers?
- Have I modified all and only the right files?  I followed existing patterns, adding the signature to a number of files in order follow the same patterns (and to cause `mpow` to be included `Owl.Mat`).  I wasn't sure whether what I did in files other than src/owl/owl_dense_ndarray_generic.ml*` was correct.
 * You have modified correct files, so no worry about it. In future, s/d/c/z and corresponding matrix files will be automatically generated from ndarray files so it will be easier. Currently we have to modify these files manually in order to reflect the changes in ndarray_generic file.

By the way, can you also add a test in [this file](https://github.com/ryanrhymes/owl/blob/master/test/unit_dense_ndarray.ml), you can run `make test` to see if it works properly. But in case you don't know how or don't have time, I can also do that later.

mars0i:
Thanks Liang.  I should be able to make those changes.  I'll look at the test file, and can probably add a test, too.

> How about negative numbers?

I'm not familiar with negative powers of matrices (except for -1, of course, but I think of that as a special-purpose notation for the inverse, rather than an exponent).  

Is the idea that e.g. `mpow x (-2)` should be `inv (mpow x 2)`?  In my initial web searches on negative powers of matrices, I didn't find anything that clearly supported this idea.

I just found [this](https://www.freemathhelp.com/forum/threads/50749-matrices-to-the-power-of-negative-numbers?s=679855ce19266484c4a49f7f77771f32&p=209027&viewfull=1#post209027) which says that  `mpow x (-2)` should be `mpow (inv x) 2`, but I'm not sure I want to take an answer on freemathhelp.com to be authoritative. :-)

mars0i:
My `mpow` function does not seem tail recursive to me, because of line 3342.  Maybe I'm not thinking about this the right way, however, because I am now unable to create a stack overflow.  For example,
```ocaml
# let m = M.(uniform 2 2);;
val m : M.mat =
         C0       C1
R0 0.292768 0.616094
R1 0.230782 0.093936

# M.mpow m 1e+308;;
- : M.mat =
   C0 C1
R0  0  0
R1  0  0
```
I'm wondering about this because there is a similar simple tail recursive algorithm [here](https://en.wikipedia.org/wiki/Exponentiation_by_squaring#Basic_method).

(There are also several more complex algorithms on that page that I would rather not investigate right now.)

Octachron:
I took the freedom to answer some linear algebra related question:

> I'm not familiar with negative powers of matrices (except for -1, of course, but I think of that as a special-purpose notation for the inverse, rather than an exponent).

As far as linear algebra is concerned, `-1` is a regular exponent for matrix exponentiation.

> Is the idea that e.g. mpow x (-2) should be inv (mpow x 2)?

Yes. More precisely, exponentiation is a group action of (ℤ,×) on GL(n,𝕂), in particular it means that 
`(M^k)^l = M^(kl)`, i.e `M^(-2) = (M^(-1))^2 =  (M^2)^(-1)`.

> (It doesn't mention non-integer powers, either, surprisingly)

Non-integer matrix powers are quite ill-defined in general, for the same reason that the logarithm is only defined up to choice of a specific branch cut in the complex plane. Often, such non-integer powers are only defined for positive semidefinite matrix since in this case all eigenvalues of the matrix are positive,
and it is thus possible to define a principal root.

 > My mpow function does not seem tail recursive to me

Your function is not tail recursive, but in the worse case, the stack grows logarithmically with the exponent, which means that any matrix with a spectral radius not extremely close to one will have overflow or underflow far before any stack overflow.

mars0i:
> I took the freedom to answer some linear algebra related question:

Thanks @Octachron.  This is very helpful.  That kind of freedom is always welcome.  Your reply makes several things clear.

>Your function is not tail recursive, but in the worse case, the stack grows logarithmically with the exponent, which means that any matrix with a spectral radius not extremely close to one will have overflow or underflow far before any stack overflow.

Good point.  Also, for an exponent that's close to `infinity`, the log of the exponent is small, so not much stack *can* be used with `float` exponents.  e.g. `log_2 (10 ** 308)` is 1024.

OK, so maybe there's a small value to rewriting the function to make it tail recursive, but given that costs would be dominated by matrix multiplication and blowing the stack is impossible in practice, it doesn't seem worth changing the function.

mars0i:
Now `mpow`, in owl_dense_ndarray_generic.ml, uses `eye`, which is defined in owl_dense_matrix_generic.ml.  `Owl_dense_matrix_generic` includes `Owl_dense_ndarray_generic`.

I don't know what the preferred solution is.  Move `eye` to owl_dense_ndarray_generic.ml?  Or move `mpow` to owl_dense_matrix_generic.ml?  I'm not sure why `mpow` should be in the ndarray module rather than in the matrix module.

mars0i:
Ah--the one-parameter `eye` is defined in owl_dense_matrix_[s,c, ... etc.].ml.  So I would get the kind of the matrix argument to `mpow` to call the two-parameter `eye`.  I think that would be necessary even if `mpow` builds the identity matrix without calling `eye`.

No urgency of course.  I'll wait for advice.

ryanrhymes:
@Octachron Florian, thank you very much for the detailed explanation 💯 @mars0i this PR is already solid enough so I will merge it and I myself can fix the zero exponent case later. 

`Linalg` module has the function `is_posdef` that can check if a matrix is positive semidefinite, but we can leave the non-integer case implementation for future.

ryanrhymes:
@mars0i OK, I moved `eye` to ndarray, also added unit test for `mpow`. Great PR, thanks again Marshall, also thank @Octachron for the professional explanation.

mars0i:
Great.  Thanks Liang.

There’s an error in the last line of my code:  r should be negated.

I was going to fix it in the next commit but wanted to keep the number of commits low.  I don’t have wifi right now, though.  I’ll submit a PR as soon as I can, unless someone else gets to it first.

Best,
Marshall

Sent from my iPhone



Sent from my iPhone
On Dec 24, 2017, at 6:47 AM, Liang Wang <notifications@github.com<mailto:notifications@github.com>> wrote:


@mars0i<https://github.com/mars0i> OK, I moved eye to ndarray, also added unit test for mpow. Great PR, thanks again Marshall, also thank @Octachron<https://github.com/octachron> for the professional explanation.

ryanrhymes:
@mars0i no problem, I will leave the fixing job to you then ;-)

PR #142, opened 2017-12-24
 fixed negative exponent case of mpow 

mars0i:
(Why does github include all of the other commits, even though they've been merged?!  I'll have to figure that out for future commits.)


ryanrhymes:
@mars0i no worry about this, there is a "Squash and merge" option in github so all your commits can be treated as a single one before merging.

PR #143, opened 2017-12-28
In docs for inv, added note about singular matrices

mars0i:
Liang--I can't squash commits.  It looks like there's something that you have to enable in order for others to be able to squash:
https://github.com/blog/2141-squash-your-commits

(I tried to squash in git following these instructions:
https://gist.github.com/patik/b8a9dc5cd356f9f6f980
It didn't work; now I have two different commits for the exact same changes, plus an irrelevant merge. :-)


ryanrhymes:
I can choose `Squash and merge` whenever accepting your PR, so it works on my side rather than yours :)

Thanks for helping me in keeping the doc updated. I will rebuild all the doc soon and upload to the server.

PR #145, opened 2017-12-28
updated version to 3.0, added note about opam bug

mars0i:


ryanrhymes:
@mars0i a better solution is using `%%VERSION%%` rather than hardcoding the version number, e.g. https://github.com/ryanrhymes/owl/blob/f46b8b47f9640a470d1f8f6c96542abfa2720c84/src/owl/owl.ml#L9

mars0i:
I like that, but it doesn't seem to work.  i.e. if I insert `%%VERSION%%` instead of `0.3.0` in README.md, it should be replaced with the current version,r ight?  But currently `%%VERSION%%` appears at the top of README.md, yet it's not getting replaced; what's displayed in github under "Owl - An OCaml Numerical Library" is just that literal string, "%%VERSION%%".

ryanrhymes:
jbuilder has a switch to allow preprocessing and replace this with current version.

PR #147, opened 2017-12-29
functorise owl_neural

bcdarwin:
Note that owl_neural_parallel is currently hardcoded to use single-precision arithmetic, hence can't be supported in the Neural.D module.


ryanrhymes:
Thanks for fixing this issue, really appreciate it 💯

PR #151, opened 2018-01-20
Additional tests for algorithmic differentiation

atthecodeface:
This update includes a new test case class for the algorithmic differentiation, running across all the scalar functions supported by algodiff, for a number of real values.

There are two operating modes for the test; the first is to run a test for a function f with a known good derivative function df individually on floats; this seems to work perfectly.

The second mode is to use the same functions f and df, and use algodiff to apply diff to Arr x_arr, generating an array of values for the derivative at each element of x_arr, and to compare this to an array of values calculated individually through df.

This second test is currently failing for specific functions f, and this is a bug in algodiff.

It fails if f is, for example, fun x -> Math.(F 1.)

It also fails if f is, for example, fun x -> Math.(x), or fun x -> Math.(x + x)

The failure is that algodiff does not derive an array when 'diff f (Arr ba)' is called.

It does work for more complex functions such as fun x -> Math.(x * x)

In particular, it fails for f=1.; f=x; f=x+x; f=2x-x; and f=-x


ryanrhymes:
This is awesome, thank you Gavin. 💯  I will have a look to see how to fix these issues.

test/unit_algodiff_ad.ml
```diff
@@ -44,9 +45,358 @@ module To_test = struct
...
+
+     df : float -> float maps x to the value of df/dx at x
+
+     x : float **)
+
+  let check_derivative_array ?eps:(eps=0.) f df x_arr =
+    M.print x_arr;
+    let df_arr   = M.mapi (fun o x -> df x) x_arr in
+    M.print df_arr;
+    let diff_arr = (diff f (Arr x_arr)) |> unpack_arr in
```
ryanrhymes: This line should be 

```
let diff_arr = M.map (fun x -> diff f (F x) |> unpack_flt) x_arr in
```

ryanrhymes:
There is one line in the test code seems incorrect, after fixing this line, all the issues are gone. So this seems a false alarm, phew :)

test/unit_algodiff_ad.ml
ryanrhymes: A further note on this: 

The reason is: if you pass in a vector to a function, based on your output (scalar or vector), it can be either jacobian or hessian if you want to differentiate a function, so it will confuse `Algodiff` if you don’t specify clearly.

PR #154, opened 2018-01-24
Add unit tests for convolution and pooling operations

jzstark:
This update includes unit tests for these operations: `avg_pool2d`, `max_pool2d`, `avg_pool2d_backward`, `max_pool2d_backward`, `conv2d`, `conv2d_backward_input`, `conv2d_backward_kernel`, and their 3d version operations. The tests for `max/avg_pool3d_backward` are not included for now since these two operations are not yet implemented in Owl.

I mainly refer to Tensorflow's tests for [conv2d](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/tests/conv2d_test.py), [conv3d](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/kernel_tests/conv_ops_3d_test.py), [pool2d](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/tests/pooling_ops_test.py), and [pool3d](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/tests/pooling_ops_3d_test.py) operations during writing these tests. The operation mode is to run a test for an operation with a known output and then compare them, with various input, kernel, stride and padding parameter. 

Most of these tests works perfectly, but four test cases fail for the`max_pool2d_backward` and `avg_pool2d_backward` operations, when the padding is set to `SAME`. The errors are due to the corner cases where part of a kernel window is in the input ndarray and the other part is zero padding. The back propagation of such window is ignored in `max_pool2d_backward`, and in `avg_pool2d_backward` the gradient is computed as `sum / window_size` rather than `sum / effective_window_size`.  

Not sure if this is a good practice but I have commented these four tests for now (see line 290 of `test/unit_pool2d.ml`). I have fixed these errors on locally implemented pooling operations. I plan to PR my code soon this week. If this method is considered slow, then a fast patch of Eigen code might be another option.


test/unit_conv2d.ml
```diff
@@ -0,0 +1,354 @@
...
+    let expected = [|7.; 2.; 16.; 5.|] in
+    verify_value test_conv2d_back_kernel [|1;2;3;1|]
+      [|2;2;1;1|] [|2;2|] SAME expected
+
+end
+
+(* the tests *)
+
+let fun_conv00 () =
+  Alcotest.(check bool) "fun_conv00" true (To_test_conv2d.fun00 ())
```
ryanrhymes: give one line space at least between functions, this improves code readability.

test/unit_conv2d.ml
```diff
@@ -0,0 +1,354 @@
...
+let fun_conv01 () =
+  Alcotest.(check bool) "fun_conv01" true (To_test_conv2d.fun01 ())
+let fun_conv02 () =
+  Alcotest.(check bool) "fun_conv02" true (To_test_conv2d.fun02 ())
+let fun_conv03 () =
+  Alcotest.(check bool) "fun_conv03" true (To_test_conv2d.fun03 ())
+let fun_conv04 () =
+  Alcotest.(check bool) "fun_conv04" true (To_test_conv2d.fun04 ())
+
```
ryanrhymes: still, give one line space to separate functions. if you want to group them into different groups, you can add inline comments e.g. (* inverse functions ... *) and so on ...

test/unit_conv2d.ml
```diff
@@ -0,0 +1,354 @@
...
+let close a b =
+  N.(a - b |> abs |> sum') < tolerance_f32
+
+let test_conv2d input_shape kernel_shape stride pad =
+  let inp    = N.sequential ~a:1. input_shape  in
+  let kernel = N.sequential ~a:1. kernel_shape in
+  N.conv2d ~padding:pad inp kernel stride
+
+let test_conv2d_back_input input_shape kernel_shape stride pad =
+  let inp     = N.sequential ~a:1. input_shape  in
```
ryanrhymes: unless you can align all the `=` in this function, it doesn't make too much sense to just align some of them, so please make the `=` follow the variable name and leave one whitespace.

test/unit_pool2d.ml
```diff
@@ -0,0 +1,302 @@
...
+  "fun_forward05", `Slow, fun_forward05;
+  "fun_forward06", `Slow, fun_forward06;
+  "fun_forward07", `Slow, fun_forward07;
+  "fun_forward08", `Slow, fun_forward08;
+  "fun_forward09", `Slow, fun_forward09;
+  "fun_forward10", `Slow, fun_forward10;
+  "fun_forward11", `Slow, fun_forward11;
+
+  "fun_max2d_back00", `Slow, fun_max2d_back00;
+  (* "fun_max2d_back01", `Slow, fun_max2d_back01;
```
ryanrhymes: what happed to these tests, why commented them out? Do they cause failure or what?

ryanrhymes: ok, I saw your explanations in the PR.

ryanrhymes: I guess TF chooses to ignore the padding part? Do other systems such as Theano do the same thing, i.e. ignoring the padding part?

jzstark: Yes, that's what TF does in its implementation. My guess is that Theano does the same, otherwise I was wondering how Keras can give the same result when switching between backends. I'll experiment with Theano/Caffe etc. to see how they deal with pooling backward operations for these cases.

ryanrhymes: OK, is it because it was me who implemented this version in Eigen? Otherwise, eigen should call the same function. But if we are about to merge your code in the near future, then it doesn't make sense to patch the existing implementation in Eigen. We only need to guarantee the C implementation is correct and can pass the test, right?

jzstark: Totally agree. These tests are mainly for the C implementation rather than Eigen. And I think you implemented the max/avg_pool2d_backward operations in Eigen.

ryanrhymes: Haha, ok, then it was my fault then .... I was in a hurry and got everything done probably in just a couple of nights ... yes, let's focus on the c implementation. it is not worth to fix the code we are about to ditch away.

test/unit_conv2d.ml
ryanrhymes: have you fixed all these line space issue?

jzstark: Just pushed the new commit.

ryanrhymes:
Cool, looks very good and solid tests, merged.

PR #155, opened 2018-01-24
update README.md about running Owl on ARM

jzstark:
The Owl container on ARM is updated to the latest code on 20 Jan. The `gsl` is no more a dependency. Also, now the users do not need to type in `eval $(opam config env)` command before starting `utop`.


ryanrhymes:
What version of Owl does Arm image use now? Is it synced with master branch?

jzstark:
The current Arm image on DockerHub uses the v0.3.0 release, but not the most up-to-date master branch yet.

ryanrhymes:
OK, I think it is worth finding a way to automate this building process. 0.3.0 already lags behind quite a lot with current development.

jzstark:
I'm very sorry. I mistook the recent update as v.0.3.0 release. The current container uses the master branch code 4 days ago, after your recent major update.

ryanrhymes:
Hah, no need to be sorry, that is good news then since the recent update is much better :)

PR #159, opened 2018-02-07
Added interfaces, changed from functors to gadts, changed some unit tests 

tptiplea:


ryanrhymes:
You probably need to create another branch without experimental and jsof_vs_bs these stuff, then merge with that clean branch before PR, so I don't have to manually remove these files every time :)

PR #165, opened 2018-02-09
Add C implementation of pooling operations

jzstark:
This PR includes:
- replace existing eigen C++ implementation of: maxpool2d, maxpool3d, avgpool2d, avgpool3d, maxpool2d_backward, avgpool2d_backward, maxpool2d_argmax
- add new operations: maxpool3d_backward, avgpool3d_backward
- add unit test of 3D backward pooling operations
- fix existing unit test of 2D backward pooling operations

Note that currently the tests will fail because 3D backward operations are not yet included in the `Ndarray_Algodiff` signature in `base`.


ryanrhymes:
can you add corresponding signature into the base then?

jzstark:
Of course, I was just not sure if I should fiddle with the `base` code. Added 3d backward ops to the signature.  I'm adding these two operations to the OCaml implementation in `owl_base_dense_ndarray_generic.ml` and will push later.

src/base/dense/owl_base_dense_ndarray_generic.ml
```diff
@@ -2090,6 +2090,9 @@ let avg_pool2d_backward padding input kernel stride output' =
   (_pool2d_backward padding input kernel stride output'
      init_pool_fun add_val_pool_fun end_pool_fun compute_grad_fun)
 
+(* TODO *)
+let max_pool3d_backward = max_pool2d_backward
```
ryanrhymes: can this pass the test actually? I think you can just migrate your C code into pure ocaml impl, right?

jzstark: That was actually a placeholder and so I commented out the corresponding unit tests for base. I've added the correct OCaml implementation now.

src/base/dense/owl_base_dense_ndarray_generic.ml
```diff
@@ -2091,6 +2091,129 @@ let avg_pool2d_backward padding input kernel stride output' =
      init_pool_fun add_val_pool_fun end_pool_fun compute_grad_fun)
 
 
+(* TODO: definitely optimise *)
+(* General function for avg_pool3d and max_pool3d *)
+let _pool3d_backward padding input kernel stride output'
+    init_pool_fun add_val_pool_fun end_pool_fun compute_grad_fun =
```
ryanrhymes: Passing higher-order functions then calling them are very expensive operations especially in this scenario because you have so many embedded loops involving a lot of computation. I understand this way of implementing thing can significantly reduce the codebase but also slows down the performance a lot. @tptiplea is currently benchmarking the base ndarray modules, also optimising the functions there. You two @jzstark @tptiplea can probably work together on this.

I will accept current version but it needs to be optimised in future.

src/base/optimise/owl_algodiff_generic.ml
```diff
@@ -1083,6 +1083,18 @@ module Make
       A.avg_pool2d_backward p a b s o
       |> pack_arr
 
+    and max_pool3d_backward p a b s o =
```
ryanrhymes: But if you only add these two functions here, how come Algodiff module can know how to differentiate `max_pool3d` and `avg_pool3d` functions?

src/owl/core/owl_tensor.ml
```diff
@@ -0,0 +1,144 @@
...
+ * Copyright (c) 2016-2017 Liang Wang <liang.wang@cl.cam.ac.uk>
+ *)
+
+(** Tensor: tensor implementation in Owl *)
+
+open Bigarray
+open Owl_types
+open Owl_dense_common_types
+
+external owl_float32_tensor_maxpool_spatial : ('a, 'b) owl_arr -> ('a, 'b) owl_arr -> int -> int -> int -> int -> int -> int -> int -> int -> int -> int -> int -> int -> int -> unit =
```
ryanrhymes: `Tensor` this term has been abused recently for obvious reasons. In Owl, we stick to ndarray which simply means n-dimensional array. I suggest you to change all these (including file names) from tensor to ndarray, so these will be consistent with the naming scheme used in Owl.

src/base/optimise/owl_algodiff_generic.ml
jzstark: Sorry I thought the forward operations are already included. I'll add them in.

src/owl/core/owl_tensor.ml
jzstark: The `Tensor` is taken from Eigen's naming format. Sure, I'll update them.

ryanrhymes: I understand that, because the Tensor implementation in Eigen is provided by Google, now we are removing our dependency on Eigen, so it is a good time to change it back to ndarray, avoid future confusion.

src/owl/core/owl_tensor_stub.c
```diff
@@ -0,0 +1,84 @@
+/*
+ * OWL - an OCaml numerical library for scientific computing
+ * Copyright (c) 2016-2018 Liang Wang <liang.wang@cl.cam.ac.uk>
+ */
+
+#include "owl_core.h"
+#include <string.h>
```
ryanrhymes: Which function needs to include `string.h`?

jzstark: `memset`, for setting an array to 0.

ryanrhymes:
Another question, I remembered that you had benchmarked all the implementations and the results showed these are faster than Eigen, right? Did you benchmark function by function, or did you benchmark one DNN application as a whole?

src/owl/core/owl_tensor_stub.c
ryanrhymes: ok, I see, thanks.

jzstark:
I was doing it for each neuron. More precisely, I use a 2-node DNN: e.g. input + avgpool2d, assuming that the processing time of input node is minimal.

ryanrhymes:
OK, make sense. Remember to keep all the results, they can be included in the paper and other writings.

src/owl/core/owl_tensor_impl.c
```diff
@@ -0,0 +1,549 @@
...
+
+          TYPE acc = INITACC;
+          int max_idx = 0;
+          int c = 0;
+          for (int a = cstart; a < cend; ++a) {
+            for (int b = rstart; b < rend; ++b) {
+              if (a >= 0 && a < input_cols &&
+                  b >= 0 && b < input_rows) {
+                int input_idx =
+                  input_idx_base + a * input_ri + b * in_channel + l;
```
ryanrhymes: In this PR, this is fine. I will find some time in discussing with you next week, this can be potentially optimised further.

src/owl/core/owl_ndarray_stub.c
```diff
@@ -0,0 +1,84 @@
...
+#undef UPDATEFN
+#undef ACCFN
+#undef INITACC
+#undef TYPE
+#undef FUN_BYTE
+#undef FUN_NATIVE
+#undef OWL_TENSOR_MAX
+
+
+#define OWL_TENSOR_AVG
```
ryanrhymes: also these? From TENSOR --> NDARRAY?

ryanrhymes:
Great, this a terrific PR! Very solid implementation and evaluation, very valuable for Owl's DNN module. 💯 👍

jzstark:
Thanks a lot!  Finally finish the pooling operations :) I'll discuss possible C code optimisations with you next week.

PR #172, opened 2018-02-14
Fixed bug

tptiplea:
This was copied from owl_dense_ndarray_generic with a small bug, but it seems the bug there was fixed.


ryanrhymes:
Thanks, I noticed that in my code and fixed.

PR #173, opened 2018-02-14
Add SqueezeNet example

jzstark:
Add the SqueezeNet DNN structure to `examples`.  The corresponding gist for inference is [here](https://gist.github.com/jzstark/c424e1d1454d58cfb9b0284ba1925a48).


ryanrhymes:
good example, thanks.

PR #175, opened 2018-02-17
Initial support for version in Zoo

jzstark:
This PR contains my current initial work on adding support to different versions of a gist. Please close it as you see fit.


src/zoo/owl_zoo_cmd.ml
```diff
@@ -113,17 +120,22 @@ let run_gist gist =
...
+
 let print_info () =
   let info =
     "Owl's Zoo System\n\n" ^
     "Usage: \n" ^
     "  owl [utop options] [script-file]\texecute an Owl script\n" ^
     "  owl -upload [gist-directory]\t\tupload code snippet to gist\n" ^
     "  owl -download [gist-id]\t\tdownload code snippet from gist\n" ^
-    "  owl -remove [gist-id]\t\t\tremove a cached gist\n" ^
+    "  owl -remove [-a] [gist-id]\t\t\tremove a cached gist\n" ^
```
ryanrhymes: `-a` means all? Do you think we need this? What if I don't add `-a`, what the command will delete then?

src/zoo/owl_zoo_cmd.mli
```diff
@@ -6,7 +6,7 @@
 
 (** {6 Manipulate gists} *)
 
-val remove_gist : string -> unit
+val remove_gist : ?all:bool -> string -> unit
```
ryanrhymes: I think the signature should be: `?vid -> string -> unit`, if vid is None, then remove the whole gist, otherwise remove one specific version, right?

What is your string then? Is it just a gist id, or is it the whole name scheme ... such as `gist | vid | ...`

jzstark: Definitely correct. Adding `-a` is only a temporary solution. I think the signature could be kept unchanged, and `remove_gist gist` parses the whole gist name (`gid/vid/pin`) to see if there is a `vid` contained.

src/zoo/owl_zoo_cmd.ml
ryanrhymes: ?

src/zoo/owl_zoo_cmd.mli
ryanrhymes: I think your coding still tangles everything together. Do you really think `remove` a zoo gist and `install/download` a zoo gist should understand the gist string in the same way?

For installing a zoo gist, you can expand the gist id as we discussed before. But for removing, I am not sure what you are doing is correct. What does `pin` mean to `remove`???

jzstark: OK I see, `remove` does not need to understand `pin`, only `gid` and optionally `vid`. My previous idea only makes this function pointlessly complex.

ryanrhymes: The point when you are designing the naming scheme, you need to look in every command need to have the same understanding or not. You can often categorise them into several groups. `remove` imo is not in the same one as `download`. 

I have been asking you to maintain a specification file, explaining the grammar and semantics of the naming and how each key command understands it. I guess you really need it, otherwise your coding is just patching rather than designing.

jzstark: Understand. Put the specification on the todo list. Thanks for helping to fix.

src/zoo/owl_zoo_cmd.ml
```diff
@@ -28,47 +29,54 @@ let preprocess script =
   tmp_script
 
 
-let remove_gist gist =
-  Owl_log.debug "owl_zoo: %s removed" gist;
-  let dir = Sys.getenv "HOME" ^ "/.owl/zoo/" ^ gist in
+(* TODO: use single gid to express "removing all versions of this gist" *)
```
ryanrhymes: I think we can agree on removing a gist id is just removing it completely from the disc. Do you agree? Because I cannot really find a convincing example for just removing one version of it. But if you have good cases to convince me, I am open to different solutions.

src/zoo/owl_zoo_cmd.ml
```diff
@@ -28,47 +29,54 @@ let preprocess script =
...
-  Sys.command cmd |> ignore
+  let ret = Sys.command cmd in
+  if ret = 0 then (
+    Owl_zoo_log.remove_log gid vid;
+    Owl_log.debug "owl_zoo: %s removed" gid
+  )
+  else Owl_log.debug "owl_zoo: Error remvoing gist %s" gid
 
 
 let upload_gist gist =
```
ryanrhymes: Does this function support updating one specific version of a gist? This indicates `upload` understands the naming scheme in the same way as `download`, right?

src/zoo/owl_zoo_cmd.ml
jzstark: The only case I can think of is removing previous versions to free disk space, but that can be more effectively achieved by `remove` all the versions and then `download` latest version. So we can agree on it.

src/zoo/owl_zoo_cmd.mli
```diff
@@ -50,3 +50,6 @@ val print_info : unit -> unit
 
 val start_toplevel : unit -> unit
 (** Start the toplevel system tailored for Owl's zoo system. *)
+
+val extend_dir : string -> string
```
ryanrhymes: Do you call this function out of the module? I think it is just  an internal function within the module, if so no need to expose it in the mli file.

src/zoo/owl_zoo_cmd.ml
jzstark: In `upload`, though the full gist name is passed in but only the gist id is used.  It does not support updating a specific version.

src/zoo/owl_zoo_cmd.mli
jzstark: Yes, this function intended to provided a way to correctly locate the files in the `~/.owl/zoo` directory in a zoo script.

src/zoo/owl_zoo_dir.ml
```diff
@@ -3,55 +3,101 @@
...
+    if (len = 1 || len = 2) then false
+    else if (len = 3) then (
+      if (List.nth lst 2 = "pin") then true else false
+    )
+    else failwith "Zoo error: Illegal parameters numbers"
+  in
+  (validate_len gid 32 "gist id"), (validate_len vid 40 "version id"), pin
+
+
+let _deploy_gist gid vid =
```
ryanrhymes: A bit confused by the name, so this deploy is just download/install zoo gist locally, it has nothing to do with the zoo depolyer we discussed before, right? If so, name probably needs to be changed to avoid ambiguity.

jzstark: Yes, this is just a ambiguous name off the top of my head. I'll change it.

src/zoo/owl_zoo_cmd.ml
ryanrhymes: Sorry, my bad. Updating one specific version does not really make sense, every time we upload a gist, we generate a new version. The reason is I saw you called `parse_gist_string` ... then why call this function if you only use gist id.

I think we have the same situation here as for `remove` ... `upload` only needs to understand gist id rather than the naming scheme ... so calling parsing function is a bit pointless ... is it, unless you do extra check there?

src/zoo/owl_zoo_cmd.ml
```diff
@@ -92,11 +100,10 @@ let show_info gist =
 
 (* format "gist/file", e.g., d7bdd62b355f906ed059f00b1270b79c/readme.md *)
 let load_file f =
```
ryanrhymes: With the new scheme, do you think this function still works properly? (because you are using subfolder, i cannot see it will work without using your extend_dir function.

src/zoo/owl_zoo_cmd.ml
jzstark: Yes, it's similar to `remove`: I assumed every command takes the same full gist name, but what this function needs is only `gid`.

src/zoo/owl_zoo_dir.ml
```diff
@@ -3,55 +3,101 @@
  * Copyright (c) 2016-2018 Liang Wang <liang.wang@cl.cam.ac.uk>
  *)
 
+let dir = Owl_zoo_log.dir
 
-let rec _extract_zoo_gist f added =
+let parse_gist_string gist =
```
ryanrhymes: a suggestion,  i think this function can be better written with pattern matching 

```ocaml
match s with
| gist :: [] -> ...
| gist :: version :: [] -> ...
| _ -> ....
```

src/zoo/owl_zoo_cmd.ml
jzstark: My bad. I'll fix this function.

src/zoo/owl_zoo_dir.ml
ryanrhymes: Also, because this is your core function, you should put expand rule here, at least briefly. For complicated functions, you need to have necessary inline doc.

src/zoo/owl_zoo_log.ml
```diff
@@ -0,0 +1,85 @@
+(*
+ * OWL - an OCaml numerical library for scientific computing
+ * Copyright (c) 2016-2017 Liang Wang <liang.wang@cl.cam.ac.uk>
+ *)
+
+let dir = Sys.getenv "HOME" ^ "/.owl/zoo/"
```
ryanrhymes: you have been using global dir everywhere in every module, spreading everywhere. If path is used so often, consider a separate module e.g. `owl_zoo_config` for accumulating this vars, and for making path such as extend_dir and etc.

src/zoo/owl_zoo_dir.ml
jzstark: Agree. Documentation is surely required for it to be used by your doc and other users.

src/zoo/owl_zoo_log.ml
```diff
@@ -0,0 +1,85 @@
...
+  let buf = Buffer.create 50 in
+  (try
+     while true do
+       Buffer.add_channel buf ic 1
+     done
+   with End_of_file -> ());
+  Unix.close_process (ic, oc);
+  Buffer.contents buf
+
+let _create_log () =
```
ryanrhymes: log often means logging everything, but i guess you are mostly dealing with versions here ... I am not sure log is a good name here, but i also ack i don't have a good suggestion yet, can think further.

src/zoo/owl_zoo_log.ml
jzstark: Thanks. I believe path should be named only once rather than using "~/.owl/zoo/" everywhere. A separate module is better solution than a global var.

src/zoo/owl_zoo_log.ml
jzstark: How about `zoo_version_db`? Or since it's already in the `~/.owl/zoo` directory, so maybe just `version_db`?

ryanrhymes: up to you, but i do find log is not very obvious for its purpose, can be short as `zoo_ver` or surreal as `zoo_keeper` :)

jzstark:
Thanks for all the feedback! I've fixed the aforementioned issues, and add a brief description of naming scheme [here](https://github.com/jzstark/worklog/blob/master/version_naming.md). There is still work to be done on test, especially about `pin`.

jzstark:
Totally agreed! I was going to say that it incurs traffic every time a user runs a zoo script, but then it turns out that it actually quite make sense to only keep one option, not only because of reducing ambiguity, but also because all versions are totally equal and there should be nothing special about the newest cached version. Maybe checking the newest cached version should only be made a default action of `latest`, for example, when there is network problem and it cannot get information from the gist server.

ryanrhymes:
Yes, I think semantically `latest` always indicates the newest version available. But in practice, you can always retrieve from local unless: 1) user manually updates the gist; 2) certain period passes (e.g. 1 week) then you force zoo to contact the server to refer the newest version.

I think we agree on this now, you can update the spec accordingly.

ryanrhymes:
@jzstark Assume `gid/remote/pin`  and `gid/remote/unpin` are gone since you only keep `latest`. Now I will talk about second issue.

I think `gid/latest/pin` is also ambiguous. If this means pin the latest version locally available, then how Zoo will understand it when the gist is pulled by another user on a different computer? This makes this name completely meaningless ...

The only solution imo is that Zoo replace `latest` with the actual vid automatically when the first time gist author uploaded one gist. If you don't want to implement this at the moment. I suggest you drop this naming.

jzstark:
Yes, agree. The same `gid/latest/pin` command could leads to different interpretations on different machines / at different time. I'll drop this naming. But I'm not sure if the first version id is a good replacement, since that version is actually the "oldest" and in many cases not ready to be used.

ryanrhymes:
> But I'm not sure if the first version id is a good replacement, since that version is actually the "oldest" and in many cases not ready to be used.

@jzstark I don't understand this. What do you mean by first version id??? This is **NOT** the first version of the gist the developer just made, it is the id that this gist refer to. I think you really need spend more time in thinking about the naming specification. I am still not very content with current state and your dev work all depend on this!

ryanrhymes:
I am closing this PR, too many issues in your design and code. I suggest that you refine the naming scheme first, submit it as next PR. When the naming scheme is solid enough, you then continue the dev and eval. Otherwise all your keep repeating the same work and going nowhere, this should have been finished long before.

PR #181, opened 2018-02-20
Add basic root finding algorithms

kanisteri:
I have added three basic root finding algorithms: bisection, false position, and ridder algorithms. Please have a look. If this is ok, I can add mli file later. 

I currently PR to the core library rather than the base library, but you can decide where to put these functions. 

Thanks.


ryanrhymes:
This is fast! The code looks very good to me. 💯

PR #182, opened 2018-02-20
Add Brent method for root finding

kanisteri:
This PR includes the implementation of Brent algorithm. I also fixed some documentation in the Stats module.

### DIFF
--- a/src/owl/jbuild
+++ b/src/owl/jbuild
@@ -161,7 +161,8 @@
     -g -O3 -Ofast
     ; Experimental switches
     ; -ffast-math may break IEEE754 semantics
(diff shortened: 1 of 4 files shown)


ryanrhymes:
Awesome! Many thanks for the Brent algorithm, also thanks for fixing the doc, really appreciate it 💯

ryanrhymes:
I went through your code, really good. I will add another function `fzero` similar to Matlab, as a hub function for those individual ones.

kanisteri:
@ryanrhymes Sure, adding the `fzero` function seems very reasonable. Thanks.

PR #183, opened 2018-02-21
Add VGG16 example

jzstark:
- Add the VGG16 DNN structure to examples. The corresponding gist for inference is [here](https://gist.github.com/jzstark/f5409c44d6444921a8ceec00e33c42c4). 
- Add link to the original paper for each DNN network structure in examples.


ryanrhymes:
That looks very good to me, thanks! 💯

PR #184, opened 2018-02-21
Helper function and unit test for Root finding

kanisteri:
In this PR, I have included a helper function for root-finding algorithms, more specifically for expanding bracket. I also included unit tests for the root-finding algorithms.


ryanrhymes:
That looks very good to me! 💯

PR #188, opened 2018-03-09
Add C implementation of convolution operations

jzstark:
This PR replaces existing Eigen C++ implementation of `conv2d`, `conv2d_backward_input`, `conv2d_backward_kernel`, `conv3d`, `conv3d_backward_input`, and `conv3d_backward_kernel` with C code, for both `float32` and `float64` precision.

About the running speed:
- forward operation: close to the original Eigen implementation [[eval figure](https://github.com/jzstark/PyFragments/blob/master/Other/conv_impl_eval/conv_impl_cmp.pdf)]
- backward_input operation: the c implementation is faster when input size is small, but when input size grows, Eigen's input can be at most 1.5 times faster than the c implementation.
- backward_kernel operation: Eigen's backward_kernel is somehow very slow compared to its backward_input, and the C implementation is about 8 times faster than it. 
- the above results only concern 2D operations, and I assume the 3D operations perform similarly.


ryanrhymes:
This is an awesome PR! 👍 Thank you Jianxin. It is very important to Owl's core because since this PR, all the operations on dense data structure fully use Owl's own native implementation, without any dependency on third-party libraries.

I am very happy to see this high-quality design, implementation, and evaluation work, well done 💯

PR #191, opened 2018-03-12
Refine the upload function in Zoo

jzstark:
This PR includes:
- fix a issue that `upload` in Zoo cannot upload binary files to gist
- a minor change of `upload_gist` function so that it returns the uploaded gist's id
- change the standard naming of the readme file to `#readme.md` to make sure it always shows on the top in the webpage


src/zoo/owl_zoo_cmd.ml
```diff
@@ -35,10 +48,11 @@ let remove_gist gist =
...
 
-let upload_gist gist =
-  Owl_log.debug "owl_zoo: %s uploading" gist;
-  let cmd = Printf.sprintf "owl_upload_gist.sh %s" gist in
-  Sys.command cmd |> ignore
+let upload_gist dir =
+  Owl_log.debug "owl_zoo: %s uploading" dir;
+  let cmd = Printf.sprintf "owl_upload_gist.sh %s" dir in
+  Sys.command cmd |> ignore;
+  _read_file (dir ^ "/gist.id") |> _strip_last_char
```
ryanrhymes: What is this for? Is this for removing the `\n` in the string? If so, this is not the right way because what if the last char is not `\n`?

ryanrhymes: Why don't you just use the `read_file` function in `Owl_utils` which already contains a `trim` parameter?

src/zoo/owl_zoo_cmd.ml
```diff
@@ -5,6 +5,19 @@
...
+  let s = Bytes.create n in
+  really_input ic s 0 n;
+  close_in ic;
+  (s) |> Bytes.to_string
+
+
+let _strip_last_char str =
+  if str = "" then "" else
+  String.sub str 0 ((String.length str) - 1)
+
```
ryanrhymes: what is the use case of this function?

src/zoo/owl_zoo_cmd.ml
```diff
@@ -5,6 +5,19 @@
 
 open Owl
 
+let _read_file f =
+  let ic = open_in f in
```
ryanrhymes: There are already `read_file` and `read_file_string` two functions in Owl_utils module. I don't see the point of making another similar function here. Please search through the code repo to make sure there are no such functions before start implementing.

ryanrhymes: But your impl is probably more efficient than the one in Owl_utils. I can move your code there later.

jzstark: Sorry for the late reply. There I just take some familiar code snippets  to finish the task at hand, and didn't realise Owl_utils has already provided it. The efficiency is definitely not a concern here. I'm updating the code.

src/zoo/owl_zoo_cmd.ml
jzstark: This function is solely for stripping the last character `\n` from the gist string that was read in from the `gist.id` file, since the string returned by my `_read_file` function always contains the `\n`.

src/zoo/owl_zoo_cmd.ml
ryanrhymes: No problem, I later realised that it is my fault, because the documentation does not include all the functions in Owl_utils module. I have migrated your code into `Owl_utils.read_file_string` and pushed to the master. So you can use that function now.

src/zoo/owl_zoo_cmd.ml
ryanrhymes: OK, then you can use `String.trim`. Or you can just use `Owl_utils.read_file ~trim:true` then take the first element in the array, this might be the best solution since `read_file` return an array of string (for each line).

ryanrhymes:
Looks very neat now! Jianxin, thanks for fixing the `upload`! 💯

jzstark:
I've removed both self-defined functions and use `read_file` instead. Thanks!

PR #195, opened 2018-03-14
Zoo: Add support for versions

jzstark:
This PR includes:
- add Zoo's support for different versions of gist, without changing the current behaviour of Zoo. The implementation follows the specification.
- add a `syscall` function to `Owl_utils`, which runs system command and returns the results.
- add an exception for Zoo.

This PR does not include `pin` function and automate update to newest version of gist after a certain period. 

I've tested the code using the existing [vision services](http://138.68.155.178/zoo.html). Please check the code and I'll revise it accordingly. Thanks!


src/zoo/owl_zoo_cmd.ml
```diff
@@ -121,12 +135,12 @@ let print_info () =
...
-    "  owl -download [gist-id]\t\tdownload code snippet from gist\n" ^
+    "  owl -download [gist-id] [ver-id]\tdownload code snippet from gist; download the latest version if ver-id not specified\n" ^
     "  owl -remove [gist-id]\t\t\tremove a cached gist\n" ^
     "  owl -update [gist-ids]\t\tupdate (all if not specified) gists\n" ^
-    "  owl -run [gist-id]\t\t\trun a self-contained gist\n" ^
-    "  owl -info [gist-ids]\t\t\tshow the basic information of a gist\n" ^
-    "  owl -list\t\t\t\tlist all the cached gists\n" ^
+    "  owl -run [gist-name]\t\t\trun a self-contained gist\n" ^
+    "  owl -info [gist-name]\t\t\tshow the basic information of a gist\n" ^
+    "  owl -list [gist-id]\t\t\tlist all cached versions of a gist; all the cached gists if not specified\n" ^
```
ryanrhymes: What is the difference between `gist-name` and `gist-id`?

jzstark: `gist-name` is the full naming scheme, including `gist-id`, `version-id`, and `pin`.

ryanrhymes:
Thanks. Remember to update the corresponding tutorial, then refine the example as a motivated showcase in the draft 👍

PR #200, opened 2018-03-22
start api doc review: owl_maths.mli

nilsbecker:
changes are only in the doc strings of math functions. please check that the markup of math vs. code vs. displayed latex math is ok.  if yes i'll stick to it in future doc edits. 

style choice: one could also s/gives/returns everywhere.

### DIFF
--- a/src/owl/maths/owl_maths.mli
+++ b/src/owl/maths/owl_maths.mli
@@ -16,172 +16,174 @@
 (** {6 Basic functions} *)
 
 val add : float -> float -> float
-(** ``add x y`` *)
+(** ``add x y`` gives x + y. *)
 
 val sub : float -> float -> float
-(** ``sub x y`` *)
+(** ``sub x y`` gives x - y. *)
 
 val mul : float -> float -> float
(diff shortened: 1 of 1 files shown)


ryanrhymes:
This looks very good to me. I recommend using `:math:` more often for inline math in restructuredtext, they look really nice. Thank you Nils 💯

ryanrhymes:
Fixed some small things, now the new doc is here: http://www.cl.cam.ac.uk/~lw525/owl/apidoc/owl_maths.html

nilsbecker:
about using `:math:`: i too find math looks nicer in properly formatted rst. i'm only afraid that this fancy formatting may make the formulas completely disappear when using odoc, or make them less readable when displaying the doc strings from merlin using MerlinDocument . are these valid concerns?

ryanrhymes:
I vaguely remember `odoc` has plan to support restructuredtext in future, but I need to double-check it with my colleagues here ...

nilsbecker:
for now, i will wait with further doc updates until a preferred way to format math has crystallized.

PR #202, opened 2018-03-28
Fix bug in Zoo

jzstark:
This PR fixes a bug that cause by different behaviours of `sed` on different machines. It also removes usage of `grep` to avoid potentially similar problems.

Another fix is for `owl -list [gid]` to show brief introduction of the cached gists.


ryanrhymes:
I've tested and now it works like a charm :)

bin/owl_list_gist.sh
```diff
@@ -12,9 +12,15 @@ fi
...
+    continue
+  fi
   ID=`basename $DIR`
-  if [ -f $DIR/readme.md ]; then
-    INFO=`head -n 1 $DIR/readme.md`
+  SUBDIR=`ls -dt $DIR/* | head -n 1 | xargs echo`
+  if [ -f $DIR/\#readme.md ]; then
+    INFO=`head -n 1 $DIR/\#readme.md`
+  elif [ -f $SUBDIR/\#readme.md ]; then
+    INFO=`head -n 1 $SUBDIR/\#readme.md`
```
ryanrhymes: Oh, I just noticed, why the head file is still here? Didn't we agree on moving this into ocaml code? Can you fix this!

Issue #204, opened 2018-04-03
Gist expansion shouldn't be into `/tmp`


mor1:
Expanding gists into `/tmp/` doesn't seem such a great idea as filenames may collide (surely?), it's harder to clean up properly, could be race conditions etc.

Better to expand into a properly created temporary subdirectory perhaps? (eg., via `mktemp`)

PR #205, opened 2018-04-09
Zoo: expand Gist into temporary sub-directory

jzstark:
Following issue #204, I add a function to make temporary directories. `mk_temp_dir pat` creates and then returns the name of a fresh temporary directory: “/tmp/patXXXXXX”, with default permissions` 0o700`. Like `Filename.temp_file`, the temporary directory is guaranteed to be different from any other directory that exists when `mk_temp_dir` was called; raises `Sys_error` if the directory could not be created. The implementation is partly inspired by [Bos](http://erratique.ch/software/bos/doc/Bos.OS.Dir.html#TYPEtmp_name_pat).


mor1:
when you say "partly inspired by", what was the inspiration / how much code overlap is there?
to put it another way, is there a reason not to simply depend on `Bos` and use it directly?

jzstark:
The error around Unix syscall in `Bos` code is very well handled. That's what I followed in my implementation. My understanding is that Owl has been trying to reduce external dependencies. Besides, of all the system operations `Bos` provides, what I need here is only a function to make temporary directories. So I think implementing it in Owl might be a better choice.  Also, `Bos` uses the ISC licence which is functionally equivalent to MIT.

PR #206, opened 2018-04-16
Discrete rv docs

mars0i:
These are some docs for the discrete RV functions.   I'd be happy to revise either content and style.  Among other things, please check that the semantics are correct.  I wasn't familiar with all of the distributions, so I looked up info about some of them and had to guess about some of the parameters.  (I did try to find the source for some of the functions, but understanding the C source code for these functions wasn't trivial for me.)


ryanrhymes:
This looks very great to me. Thank you Marshall. 👍 💯

PR #208, opened 2018-04-17
Topkging

mor1:
Tidy up packaging; `make release` should now DTRT.


ryanrhymes:
Great! We can just `make release` from now on. Yay 💯

dbuenzli:
Juste note that since `owl` uses `dune`. There's now https://github.com/samoht/dune-release which should offer the `topkg` release experience without the boiler plate `topkg` needs to support older build systems.

Issue #221, opened 2018-04-20
Make gist parameter optional in Owl_zoo_cmd.load_file


ryanrhymes:
@jzstark Can we make the gist parameter optional `Owl_zoo_cmd.load_file`
? If not passed in, the load the file from current gist, the reason is that if a gist script want to load something from its own gist folder, it is difficult to what will be the gist id since it is not allocated yet.

ryanrhymes:
@jzstark thanks for fixing this 😄

PR #222, opened 2018-04-20
Make gist parameter optional in Owl_zoo_cmd.load_file

jzstark:
To address #221, this PR modifies the `load_file` function.  An example is shown in [this gist](https://gist.github.com/jzstark/8fc90f146fa81e71459385d149f3445d). A user can use `load_file ~gist f` to specify a gist as before. When the `~gist` option is not used:
- if the script is executed locally in, e.g., `utop`, then `load_file f` will try to find file `f` in local directory;
- if the script is already uploaded to Gist and run by using `#zoo`, then `load_file f` will try to find file `f` in `~/.owl/zoo/gid/vid`. Users do not need to know the gist id of the script's own Gist.


ryanrhymes:
@jzstark You may also want to sync with my master branch, I have fixed many broken things due to changing the sub library names.

jzstark:
Thanks a lot for helping to fix these library names! The code works fine on my machine.

PR #223, opened 2018-04-20
Add sum_reduce operation to ndarray

jzstark:
This PR adds the `sum_reduce` operation, which sums an ndarray along multiple axes. It's implementation is close to that of `sum`.


ryanrhymes:
Cool, can you add at least one unit test for sum_reduce?

Issue #210, opened 2018-04-17
A bug in the Normalisation neuron


jzstark:
On the forum @royson mentioned an `Exception: Owl_exception.NOT_BROADCASTABLE` error caused by using `normalisation` in a neural network. Here is his script to reproduce it:

```ocaml
#require "owl_top";;
open Owl;;
open Neural.S;;
open Neural.S.Graph;;
open Algodiff.S;;

let make_network input_shape =
  input input_shape
  |> normalisation ~decay:0.9 (* normalisation layer causes the exception below. *)
  |> lambda (fun x -> Maths.(x / F 256.))
  |> conv2d [|5;5;1;32|] [|1;1|] ~act_typ:Activation.Relu
  |> max_pool2d [|2;2|] [|2;2|]
  |> dropout 0.1
  |> fully_connected 1024 ~act_typ:Activation.Relu
  |> linear 10 ~act_typ:Activation.Softmax
  |> get_network in

let network =
  let x, _, y = Dataset.load_mnist_train_data_arr () in
  let network = make_network [|28;28;1|] in
  Graph.print network;
  let params = Params.config
    ~batch:(Batch.Mini 100) ~learning_rate:(Learning_Rate.Adagrad 0.005) 0.01
  in
  Graph.train ~params network x y |> ignore;
  network in

let imgs, _, labels = Dataset.load_mnist_test_data () in
let m = Dense.Matrix.S.row_num imgs in
let imgs = Dense.Ndarray.S.reshape imgs [|m;28;28;1|] in
Graph.model network imgs;; (* <-- Exception occur here *)
``` 

After some investigation, I find out that it is caused by [this line](https://github.com/owlbarn/owl/blob/master/src/base/neural/owl_neural_neuron.ml#L2049) in the source code:
```ocaml
let mu' = Maths.(a * (sum ~axis:l.axis x))
``` 

Here we suppose input `x` has shape `[|a; b; c; d|]`, and `axis=3`, so the shape of `mu'` will be `[|a; b; c; 1|]`, while actually the expected shape of it should be `[|1; 1; 1; d|]` (see [doc](https://www.tensorflow.org/api_docs/python/tf/nn/batch_normalization) from Tensorflow).   The shape mismatch therefore leads to that broadcast-related error. 

As comparison, here is a brief description of how `mu` and `var` parameters are computed during training in Keras:
```python
## Users specify axis and pass input data into
## keras.layers.BatchNormalization
axis = 3
[...]

PR #224, opened 2018-04-22
Fix normalisation neuron with `sum_reduce`

jzstark:
This PR fixes the #210 issue. It contains:
- Adding the `sum_reduce` operation to `Algodiff.*.Maths`. 
    - It includes a `repeat_mul` function for its backward operation, which performs `repeat` on multiple axis. For now this function is added in `Algodiff` instead of `Ndarray` layer.
- Fix the normalisation neuron with `sum_reduce`.
- Adding one test case for `sum_reduce` in unittest for ndarray.
- Adding `sum_reduce` operation in `Owl-base`. The `make test` will complain about signature mismatch if it is not implemented in the base library.


src/base/dense/owl_base_dense_ndarray_generic.ml
```diff
@@ -763,6 +763,50 @@ let max' varr = (_fold_left (Pervasives.max) Pervasives.min_float varr)
 let sum' varr = (_fold_left (+.) 0. varr)
 
 
+(** Folding along a specified axis, aka reduction.
+    m: number of slices; n: x's slice size; o: x' strides, also y's slice size;
+    x: source; y: shape of destination. Note that o <= n. *)
+let owl_sum_along m n o x ys =
```
ryanrhymes: Folding along specified axes is generic operation on Ndarray. You should pass in higher-order function to make this operator generic to other number types (such as complex) and different operations such as (such as min, max, sub, mul, and etc.).

I am also curious why you don't reuse the the existing sum operation, is it because it is too slow for calling multiple times if you need to reduce on multiple axes.

src/base/dense/owl_base_dense_ndarray_generic.ml
```diff
@@ -763,6 +763,50 @@ let max' varr = (_fold_left (Pervasives.max) Pervasives.min_float varr)
...
+          let m, n, o, s = reduce_params i !y in
+          y := owl_sum_along m n o !y s
+        )
+      done;
+      !y
+    )
+  | None   -> (
+      let y = sum' x in
+      let z = zeros (kind x) (Array.make (num_dims x) 1) in
+      Genarray.set z (Array.make (num_dims x) 0) y;
```
ryanrhymes: Why don't you just call `create (kind x) (Array.make (num_dims x) 1) y` to avoid creating unnecessary intermediate variables?

src/base/dense/owl_base_dense_ndarray_generic.mli
```diff
@@ -317,6 +317,10 @@ val sum : ?axis:int -> (float, 'b) t -> (float, 'b) t
 val sum' : (float, 'b) t -> float
 (** Refer to :doc:`owl_dense_ndarray_generic` *)
 
+(* TODO: change float to 'a *)
```
ryanrhymes: Please make this function generic, refer to my previous comment.

src/base/optimise/owl_algodiff_generic.ml
```diff
@@ -202,6 +203,18 @@ module Make
     | Arr x -> Arr A.(repeat ?axis x reps)
     | _     -> failwith "error: AD.repeat"
 
+  let repeat_mul axis x reps =
```
ryanrhymes: If my understanding is correct, this function is meant to `repeat` on multiple axes given a ndarray, right? If so, we should probably improve the API rather than adding another adhoc function. 

Please have a look at tile function, if we make the repeat function interface similar to tile, do you think this can solve the problem?

ryanrhymes: If so, please open an issue to suggest improving repeat function API asap, I can help you on this function.

src/base/dense/owl_base_dense_ndarray_generic.ml
jzstark: Yes, one of the concerns is speed. I saw the comment of the existing sum operation about its possible efficiency issue. Calling it multiple times might be quite slow. 

The other is that the `sum` in base lib reduce the input data by one dimension, and I'm sure if that's the intended behaviour of this function in base. For example:
```ocaml
module N = Owl_base_dense_ndarray.S
let x1 = N.sequential [|2;3;4|]
let y1 = N.sum ~axis:2 x1
let s1 = N.shape y1 (* s1 = [|2;3|] *)

module N = Owl_dense_ndarray.S
let x2 = N.sequential [|2;3;4|]
let y2 = N.sum ~axis:2 x2
let s2 = N.shape y2 (* s2 = [|2;3;1|] *)
```

Considering both factors, I decide to follow the implementation in `owl_dense_ndarray_generic.ml` using `owl_sum_along`.

src/base/optimise/owl_algodiff_generic.ml
jzstark: Thanks! That is exactly what this function is meant to do. I've renamed it to `repeat_axes` and added a TODO comment.

src/base/dense/owl_base_dense_ndarray_generic.ml
ryanrhymes: I see. Yes, folding along one dimension is common use case. OK, then please try to make the function generic, refer to owl_base_dense_common.ml

Ping me on skype if you need help on this.

src/base/dense/owl_base_dense_ndarray_generic.mli
jzstark: I've just updated the `owl_sum_along`, `sum'`, and `sum_reduce` functions to make them generic in [this commit](https://github.com/owlbarn/owl/pull/224/commits/1c2b309a1f057cb6b463b9b594d48cfe24a84c3c). Please take a look to see if that's a correct way to do it.

src/base/dense/owl_base_dense_ndarray_generic.ml
```diff
@@ -760,7 +760,47 @@ let max' varr = (_fold_left (Pervasives.max) Pervasives.min_float varr)
...
-let sum' varr = (_fold_left (+.) 0. varr)
+let sum' varr =
+  let _kind = kind varr in
+  _fold_left (Owl_base_dense_common._add_elt _kind) (Owl_const.zero _kind) varr
+
+
+(** Folding along a specified axis, aka reduction.
+    m: number of slices; n: x's slice size; o: x' strides, also y's slice size;
+    x: source; y: shape of destination. Note that o <= n. *)
+let owl_sum_along add_fun m n o x ys =
```
ryanrhymes: OK, I also meant that this owl_sum_along can be generalised as fold_along, which allows you to plug in not only sum, but aslo min, max, prod, and etc., so that you can further implement min_reduce, max_reduce in future if necessary.

src/base/dense/owl_base_dense_ndarray_generic.ml
```diff
@@ -760,7 +760,47 @@ let max' varr = (_fold_left (Pervasives.max) Pervasives.min_float varr)
 
 
 (* Sum of all elements *)
-let sum' varr = (_fold_left (+.) 0. varr)
+let sum' varr =
```
ryanrhymes: This looks good to me. Can you also add some notes (TODO) here to remind us to further revise other functions such as min, max, and etc.

src/base/dense/owl_base_dense_ndarray_generic.ml
jzstark: Oh I see. I think we can pass it any function of type `a'->a'->a'` to it to conduct fold operations.  I renamed this function and add TODO notes.

Issue #225, opened 2018-04-23
Bug in normalisation neuron?


ryanrhymes:
@jzstark With the recent PR #224 , the `example/cifar10_vgg.ml` is broken. Can you investigate this issue further to make sure that the normalisation neuron is really correct.

ryanrhymes:
@jzstark If I change the learning rate from Adagrad 0.005 to Adagrad 0.001, it does not return `nan` any more, but the training does not seem to converge.

jzstark:
Thanks! About using division instead multiplication --  I was trying to use multiplication at first, but then the `nan` problem was still there. I think maybe it's because by flipping the numerator and denominator, the calculate `a` would become quite small (`1 / (100 * 32 * 32) = 9.8E-6` in CIFAR example's case) so there is the float number precision issue.

PR #226, opened 2018-04-23
 Fix bug in normalisation neuron

jzstark:
This PR fix the error introduced in my previous one as shown in #225. I'm very sorry for didn't checking the code thoroughly. For future reference, I added links to the [original paper](https://arxiv.org/abs/1502.03167) on Batch Normalisation and the [implementation](https://github.com/keras-team/keras/blob/master/keras/layers/normalization.py#L130) in Keras.


jzstark:
I just noticed that after commit [9b9ad](https://github.com/owlbarn/owl/commit/9b9ad77b5fd01f7a714f30018e89b056396e1d51) Travis fails with message:
```
# File "src/owl/dense/owl_dense_ndarray_generic.ml", line 3604, characters 11-34:
# Error: Unbound value Owl_utils.Array.replace
```
Once I copy this `replace` function into `owl_dense_ndarray_generic.ml` and call it instead of `Owl_utils.Array.replace`, the Travis test is successful again.  I'm still trying to figure out what may have caused this issue.

ryanrhymes:
Is it because Owl core tries to download Owl_base from OPAM now rather than using the one in the source tree?  @mor1

src/base/neural/owl_neural_neuron.ml
```diff
@@ -2043,14 +2043,16 @@ module Make
...
-      let a = F (1. /. float_of_int (shape x).(l.axis)) in
       if l.training = true then (
+        let a = F (float_of_int ((numel x) / (shape x).(l.axis))) in
         let s = Owl_utils_array.range 0 (Array.length l.in_shape) in
         let s = List.filter (fun x -> x != l.axis) (Array.to_list s)
           |> Array.of_list in
-        let mu' = Maths.(a * (sum_reduce ~axis:s x)) in
-        let var' = Maths.(a * (sum_reduce ~axis:s (x * x))) in
+        let mu' = Maths.((sum_reduce ~axis:s x) / a) in
+        let var' = Maths.((sum_reduce ~axis:s (x * x)) / a) in
```
ryanrhymes: Can we change division to multiplication, by flipping the numerator and denominator when you calculate a? Because multiplication is cheaper than division, also for backprop phase.

Just suggesting ...

src/base/neural/owl_neural_neuron.ml
```diff
@@ -2043,14 +2043,16 @@ module Make
       mkpri l |> Array.map copy_primal' |> update l';
       l'
 
+    (* Reference: https://arxiv.org/abs/1502.03167.
+       Implementaton in Keras: https://bit.ly/2vBsvgI.*)
     let run x l =
-      let a = F (1. /. float_of_int (shape x).(l.axis)) in
       if l.training = true then (
+        let a = F (float_of_int ((numel x) / (shape x).(l.axis))) in
```
ryanrhymes: Are you sure this is correct? Shouldn't it be `F ((float_of_int (numel x)) / (float_of_int (shape x).(l.axis)))` ... please double check the reference.

And the link to the original paper is still missing.

PR #233, opened 2018-04-28
Small Change to Readme - Homebrew + git

mcsosa121:
Update readme to add additional git option. More importantly, "homebrew/science/" is now depreciated and many of the modules have just been added to the core tap. So I updated the command to reflect the new location.


ryanrhymes:
Good catch! Thank you Mike.

Issue #231, opened 2018-04-25
Extension to Zoo naming scheme


ryanrhymes:
@jzstark We had this discussion before about when to refresh local cached Zoo gists. This turns out to be very useful whenever if I need to edit some gists online then try it out immediately on local machine, I always need to run `owl -update gist_id` first.

I think we can extend the current naming scheme in Zoo by adding a field to specify how the gist should be refreshed, e.g. FREQ=Xhour/days/..., ALWAYS ... Then I can write something like

```ocaml
#zoo "6f28d54e69d1a19c1819f52c5b16c1a1/?refresh=always"
```

Does this make sense?

jzstark:
Sure! I think we can actully replace the current "latest" field with this "refresh" field to indicate the time-tolerance of a gist: "gid / refresh". By setting refresh to `0` or `always`, it means always updating the gist regardless of local cache; to `infinity` or `never` (default), it means the current "latest" setting -- always using a local gist whenever possible; and then what sits in between is `Xmins/hrs/days`. I've put this on my TODO.

ryanrhymes:
Good, glad to know it helps to address the ambiguity issue we discussed before. 0 to inf sounds good to me, we can discuss what format to represent time tomorrow.

PR #242, opened 2018-05-04
Revise Zoo naming scheme

jzstark:
This PR tries to address issue #231. It updates the previous `gid/[vid|latest]` naming to `gid?vid=xxx?tol=xxx`, where `tol` is a time value (in seconds) to indicate a gist's tolerance for the time it exists on the local cache. This change is compatible with the existing usage of zoo that in the form of `#zoo gid`. One thing that might to be done is to support richer time representation such as `3h2m5s` instead of just seconds.


src/zoo/owl_zoo_ver.ml
```diff
@@ -61,13 +61,14 @@ let get_remote_vid (gid : string) =
 
 (** Get the latest version downloaded on local machine; if the local version is
 not found on record, get the newest vid from Gist server. *)
-let get_latest_vid (gid : string) =
-  let v, _, _, miss_flag = get_value gid in
-  if (miss_flag == false) then (
+let get_latest_vid (gid : string) (tol : float)=
```
ryanrhymes: can you be a bit careful about the format? Try to avoid letting others fix these small issues for you all the time. Add space between ) and = please.

ryanrhymes:
I tried `#zoo "217ef87bc36845c4e78e398d52bc4c5b?tol=50"`, but it does not seem working properly, it always re-downloads the gist ... did I miss anything here?

jzstark:
Currently the intended behaviour is like this -- Suppose initially gist `gid` does not exist on local cache, then running `#zoo gid?tol=50` will download it at time `t1`. This time information will only be updated when a newer version is *actually downloaded* to the local cache. Now suppose `t2 = t1 + 51`, `t3 = t1 + 52`, and the gist keeps unchanged on the server, then running `zoo` at `t2` and `t3` will both trigger version information retrieval from Gist server (and thus some `curl` information will be shown, but it is not downloading the gist itself), even though `t3` is only one second later than `t2`, because at `t2` the local cache is not changed. 

But I agree that it makes more sense than this current behaviour to refresh the time at every request even though nothing is downloaded.  I'll fix it soon.

ryanrhymes:
@___@  try not make intuitively simple thing so complicated if possible .... I suggest you fix it.

ryanrhymes:
OK, I finally understood. Make sense but a bit tricky for me to figure it out in the first place ... well, you can decide whether to change it or not :)

Issue #241, opened 2018-05-04
sum function in base is inconsistent with that in core


ryanrhymes:
@jzstark @tptiplea The `sum` function in base.ndarray returns inconsistent result comparing to that in core. E.g. ``x`` has shape [2;3;4], `sum axis:0 x` returns an ndarray of [3;4] whereas the one in core returns an ndarray of [|1;3;4|]. 

Base.sum should not shrink the dimension. Please can you fix this?

ryanrhymes:
Thank @jzstark for fixing this issue so quickly! 👍

PR #243, opened 2018-05-04
Fix `sum` inconsistency in base 

jzstark:
This PR fixes the #241 issue, where sum function in `base` is inconsistent with that in core. The `sum` function is re-implemented with the `fold_along` operation. Also, `sum_reduce` is slightly modified following the recent change of `sum_reduce` in `core` to reduce redundant operations.


ryanrhymes:
mort suggested `rebase`, did you use that?

jzstark:
Sorry not yet. I tried once but ended up messing the code base. I'll figure it out for the next PR.

PR #248, opened 2018-05-14
Add transpose_conv2d neuron 

jzstark:
This PR adds the `transpose_conv2d` neuron. So far It works fine in the fast neural transfer application. I'll further add implementation of the three convolution operations in `base` library and more corner unit test cases in next PRs soon.


ryanrhymes:
Looks very good to me. Looking forward to your FST demo! 💯

PR #260, opened 2018-05-23
Call to index_1d_nd in init_nd takes stride, not shape.

paurkedal:


ryanrhymes:
Good catch! Thanks for fixing it, also really appreciate the professional unit test! 👍

PR #264, opened 2018-05-30
Non-uniform histogram

nilsbecker:
right now, i'm looking to get feedback on the api and implementation. tests are missing and the current histogram function still has to be removed. also there might be a licensing issure with the binary search function which i copied from containers and modified a bit. c-cube proposes copy-paste as one way to use his libraries, i just don't know if doing this has consequences for owl.?!


nilsbecker:
another thing that would be nice to have is the possibility to do weighted histograms. this would require a different return type, with a float array instead of or in addition to integer counts. therefore i think weighted histogram should be a different function.

ryanrhymes:
Thank you @nilsbecker , the PR looks very good to me.

- I think it is sufficient to keep the reference in the code to ACK the original implementation.
- There are two options: we can either keep both histogram functions, by renaming my original implementation to something like `histogram_sorted`. The reason is because sometimes we may already have a sorted array so this step can be saved. 
- Another potentially better option is introducing an optional parameter called `sorted` in your implementation, to indicate whether the passed in array is sorted or not. In this case, I think my original implementation can be removed.
- Weighted histogram can be a separate function e.g. `histograms_weighted` in the beginning imo. If it turns out it won't make the current API too complicated, we can also similarly using an optional parameter `weights` in `histograms` to provide a unified API.
- Does it make sense to rename ` Edge` to `Interval`?
- Does it make sense to rename `fractions_of_counts ` to `normalise`? It is a common operation imo, although normalise unfortunately has many meanings.

Yes, we do need unit tests since owl's code base has been growing very fast. Really appreciate if you can add some. Let me know your opinions on the comments above, thanks.

nilsbecker:
hi, 

- ok about the acknowledgment. maybe i could make it a bit more precise.

- i think there are four variations of histogram implementation that are of interest here:
   a histogram for uniform bins using truncation of floats
   b histogram for arbitrary bins using binary search
   c histogram of sorted data. 
   d weighted version of any of the above

the previous implementation does only a. my version does a or b depending on the bin specification. it's effectively the same as the original version in the case of uniform bins given with the `` `N`` constructor. 

i added a comment speculating about c but have not implemented anything. what i had in mind is different from the original implementation which does not exploit/depend on sorted data. i have not seen a specific histogram for sorted data but that could come in handy and histogram_sorted may be a nice addition.

d would require a different return type with floats representing summed weights. therefore i agree that it's natural to make that a different function. the only problem there is that then the different histogram versions discussed above should each get a weighted counterpart, then we would have a total of six 1D histogram functions. one could probably share code for some of the internals. 
 
- renaming `` `Edges`` to `` `Intervals``. i can't say i like `` `Edges`` very much but i think it's a bit more precise than `` `Intervals`` since the entries of the array are really the boundaries between bins. one could also just say `` `Bins`` and be clear in the docstring that these are bin boundaries?
- renaming to ``normalise` could work. then maybe the other function would be ``normalise_density`` ?

nilsbecker:
@ryanrhymes i saw your comment about the record as return type. i can do that too, maybe it allows unifying weighted and non-weighted. would you then prefer a submodule or just a histogram type flat in the Stats module?

nilsbecker:
about unit tests: where do they need to go? do i have to register them somewhere? what do i need to know to use them?

ryanrhymes:
OK, let's go with a record type without submodule for the time being, just like those hypothesis testing functions. I think returning a record type can help in unifying the weighted and non-weighted version as you said.

The right place for the unit tests should be in this module imo https://github.com/owlbarn/owl/blob/master/test/unit_stats.ml 

Test module uses https://github.com/mirage/alcotest , it is quite easy to use. The file I referred above should give you enough information, it is quite a template, you only need to add three functions for each test then edit the test_set array. `make test` will compile all the test run through them. Let me know in case you need any help on this.

nilsbecker:
weighted, unweighted and sorted histograms.

The new version now uses a record type for the histogram. the record is a single histogram type, with optional fields for the weighted counts and for normalized versions of the counts. 

the precise api is up for debate. for instance one could remove the normalized counts and density from the record and calculate those as bare float arrays instead. one could also go the other way and consolidate the weighted and and unweighted histogram functions into one function with an optional weights and normalization argument and calculate all in one go.

i briefly toyed with the idea of a GADT for the histogram type. it would be parametrized by being weighted/unweighted  and normalized/unnormalized. in principle it would then be possible to give the weighting and normalizing options as tags to the histogram function and get out a result which is a histogram with known normalization and weighting properties. but then it seemed too fancy so i did not try it.

i'm not sure about the fact that the weighted version calculates both counts and weighted counts. the counts seemed nice to have in addition but i don't know if there is an noticeable speed penalty.

the sorted histograms are only lightly tested; it would make sense to use unsafe array updates since they are purely for speed anyway but this would require being more sure that nothing goes wrong. generally tests are still missing. i'm again looking for feedback on the api.

ryanrhymes:
The PR looks very good to me now. I agree that GADT is a bit overkill here. Owl overall does not use it much except incorporating it from Bigarray itself to support polymorphism on different number types.

There is tradeoff between the API complexity and number of APIs we expose to users. I personally prefer more unified API so the programmer won't be overwhelmed by excessive number of functions they need to remember. IMO, weighted and unweighted can be unified by an optional parameter. That is also generally applied to mean, var, and other functions in Stats module, which don't support weighted version at the moment.

However, we can add these unified version in the next PR, since they will be built atop of the currently PRed functions anyway. I will merge, any further tuning and polishing of these functions are welcome. Can you also add corresponding unit tests in `test/unit_stats.ml`, many thanks!

src/owl/stats/owl_stats.ml
```diff
@@ -175,9 +175,32 @@ let max = Owl_base_stats.max
 let minmax = Owl_base_stats.minmax
 
 
+type histogram = Owl_base_stats.histogram = {
```
ryanrhymes: Can we just `type histogram = Owl_base_stats.histogram` rather than redefining every field since they are already defined in owl_base_stats .... or I missed something so compiler complains?

nilsbecker: i tried and the compiler complained... maybe there is a way but i did not find it!

nilsbecker:
ok, so i will prepare a new pr with unified histogram functions -- they will replace the current public facing ones. i will also add unit tests then. i will keep the current histogram type unless anything else comes up.

ryanrhymes:
Since there is a specific `histogram` type now, you can also consider to implement and register a pretty printer in https://github.com/owlbarn/owl/blob/master/src/top/owl_top.ml  so it will print out some useful information when using in the toplevel

nilsbecker:
yes, a printer could be useful. it gets a bit unwieldy, especially since long arrays are not truncated in pretty-printing by default. actually, i was wondering if there is going to be a vector type which will supersede the use of 1d arrays here, in the future? when thinking about n-dimensional histograms, it's clear that an Ndarray output would be required anyway.

nilsbecker:
is it possible to run just a subset of tests? this would make life easier when testing the histogram unit tests...

jzstark:
@nilsbecker  one simple way to run a subset of test is to comment out unnecessary lines in the `test_runner.ml` file.

PR #266, opened 2018-06-01
Add append option to save_txt

mars0i:
I added parameter `?(append=false)` to `save_txt`.  When `true`, append causes a new write into an existing file to append to it rather than overwriting the file.

I thought about doing something more complicated that would allow the user to set the Unix file open flags individually.  That would allow a user to specify that `save_txt` should fail if the file exists.  I'm not sure there's anything else that would really be useful in the file open flags.  My feeling is that allowing the user the choice to append rather than overwrite is a useful, simple, function that could be used often, but if the user wants to fail if the file exists, they should handle that case some other way, by checking whether the file exists before `save_txt` is called.

Similarly, we could allow the user to pass a file permission value to `save_txt`, but this just seems like too much for a simple function.  The user can go back and change the permissions separately.

So I have only added `~append` as a boolean that determines whether to append or overwrite.

### DIFF
--- a/src/owl/dense/owl_dense_matrix_c.mli
+++ b/src/owl/dense/owl_dense_matrix_c.mli
@@ -419,7 +419,7 @@ val save : mat -> string -> unit
 
 val load : string -> mat
 
-val save_txt : ?sep:string -> mat -> string -> unit
+val save_txt : ?sep:string -> ?append:bool -> mat -> string -> unit
 
 val load_txt : ?sep:string -> string -> mat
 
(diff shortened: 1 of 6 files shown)


mars0i:
Oh, I meant to mention that I didn't do anything to the `save_txt` functions in the `*nlp*` source files.  I'm not sure what those are exactly.  Also, there weren't any tests for `save_txt` and I didn't add any.

ryanrhymes:
Thank you Marhsall, the code looks very good to me. Yes, the save_text function in Ndarray module is meant to be easy to use so `append` is a good solution. If the programmer wants better control, they can choose to use the functions in `Owl_io` or those IO apis directly.

Yes, NLP is quite an independent module, the functions there are doing something slightly different, but we can ignore this module at the moment :)

PR #268, opened 2018-06-04
updated owl_maths.mli with :math: annotations

nilsbecker:
this should make the math display in html generated docs look nicer


ryanrhymes:
Many thanks, the doc has been rebuilt: http://ocaml.xyz/apidoc/owl_maths.html

The new doc looks much better 👍

PR #270, opened 2018-06-06
1D histogram improvements

nilsbecker:
This now adds unit tests and a pretty-printer.

Some bugs have been ironed out, and some refactoring has made the code a little nicer. 

about the included modified piece of containers, see here:
https://github.com/c-cube/ocaml-containers/blob/master/LICENSE
it might be that this text has to be included? not sure.


ryanrhymes:
Looks all good to me. Thanks for adding the tests and polishing the APIs. 👍

PR #272, opened 2018-06-14
Add dilated convolution and update related code format 

jzstark:
This PR contains:

- Add dilated convolution operations, neurons, and unit test;
- Add mutable operations for existing convolution operations;
- Move `owl_utils_conv` functions and par of `owl_utils_ndarray` functions to `owl_utils_infer_shape`; shape inference functions of dilated convolutions are also added;
- Adjust the order of convolution functions in `mli` files;
- Add unit test for previously added `conv_mec` and `conv_naive` implementation.

One thing that is not done yet is the name of functions from `owl_utils_conv`. Current I keep them unchanged as e.g. `calc_conv2d_output_shape`. Maybe later we should find some proper renaming rules for them.


ryanrhymes:
Awesome PR!

PR #277, opened 2018-06-27
Change interface of repeat function

jzstark:
This PR re-implement `repeat` in `core`, and change its interface to be the same as `tile`, so that repeat can be performed on multiple dimensions rather than one axis.


src/base/core/owl_computation_eval.ml
```diff
@@ -53,7 +53,7 @@ module Make (A : Ndarray_Mutable) = struct
         | Reshape shape                               -> _eval_map_01 x (fun ~out x -> A.reshape_ ~out x.(0))
         | Reverse                                     -> _eval_map_01 x (fun ~out x -> A.reverse_ ~out x.(0))
         | Tile repeats                                -> _eval_map_00 x (fun x -> A.tile x.(0) repeats)
-        | Repeat (axis, repeats)                      -> _eval_map_00 x (fun x -> A.repeat ~axis x.(0) repeats)
+        | Repeat repeats                              -> _eval_map_00 x (fun x -> A.repeat x.(0) repeats)
```
ryanrhymes: this is not correct. but since you are not very familiar with computation graph, so it is fine, i will fix it later.

ryanrhymes:
Close this PR because the stack grows too big in the worst case.

PR #290, opened 2018-08-22
use explict labels in cblas and lapack modules

struktured:
This PR removes the need for `[@warning "-6"]` in `owl_lapacke.ml` and `owl_cblase_basic.ml` by explicitly labeling all the arguments. While it has no immediate value beyond readability, it may reduce friction when maintaining these modules in the future.


ryanrhymes:
@struktured new owl_cblas_basic.ml fails four unit tests. Can you dig a bit deeper to fix this?

I temporarily reverted the cblas basic, the lapacke one works fine. Thanks.

screen shot 2018-08-22 at 12 23 05

PR #291, opened 2018-08-22
use explicit labels in owl_cblas_basic

struktured:
This PR corrects a copy/paste bug discovered when running tests with #290.


ryanrhymes:
Great, many thanks for fixing this.

PR #294, opened 2018-09-01
Add UpSampling2D neuron

jzstark:
This PR adds the upsampling2D operations, neuron, and unittest. I also try to add it into the computation graph.  I plan to add the 1D and 3D implementation later.


ryanrhymes:
Awesome PR, thank you Jianxin!

PR #296, opened 2018-09-04
tukey fences implementation

struktured:
This PR adds a function to compute [tukey fences](https://en.wikipedia.org/wiki/Outlier#Tukey's_fences):

`val tukey_fences : ?k:float -> float array -> float * float`

Given an array of floats, it determines the lower and upper bounds for non outlier values.


ryanrhymes:
Thank you Carmelo.

PR #297, opened 2018-09-05
dune: Use 1.1.x simplifications

mc10:
This is untested, but I think it's okay.


ryanrhymes:
Thanks, I've tested the PR but it fails to build on my machine.

mc10:
Closing for now; I'll make a new PR once I get the build to work.

PR #298, opened 2018-09-05
travis: Use the original ocaml-ci-scripts script

mc10:
This should fix the current Travis build errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,10 @@ sudo: false
 services: docker
 
 install:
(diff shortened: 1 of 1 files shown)


mor1:
LGTM, thanks! 

FWIW this won't fix the Travis build errors because they are due to various Blas/LAPACKE/etc build nonsense -- different packaging on different distros builds and then installs libraries in different places with, in some cases, different symbols :(  But since my PR to quieten git was accepted, we should be back on the mainline rather than my fork, so it's still an improvement.

PR #301, opened 2018-09-09
Improve numerical accuracy in stats distributions

ejaasaari:
Improves numerical accuracy of the statistical distributions to help dealing with small probabilities. Previously for example

`Owl.Stats.binomial_logpdf 1 ~p:0.5 ~n:2000`

would return -infinity, whereas now it returns `-1378.69345866034632`.


ryanrhymes:
This PR indeed improves the numerical stability of Owl's statistical functions. Many thanks Elias 💯

PR #303, opened 2018-09-10
kernel density estimator, ported from pareto

struktured:
This PR introduces a simple kernel density estimator which is ported from `pareto`'s [sample module](https://github.com/superbobry/pareto/blob/master/lib/sample.ml#L242). It is a direct port- there was no refactoring done during the migration, only changes from `gsl` invocations to `owl`.

The implementation has been (tentatively) placed in `owl/stats/owl_stats_kde.ml[i]`, but I would appreciate input as to where it best belongs in `owl`.

In `pareto`, the estimate pdf function accepts an optional variant to denote the type of kernel- personally I would prefer something like a functor or first class module to indicate / depict kernels, but am curious to know other opinions on the matter.


src/base/stats/owl_base_stats.mli
```diff
@@ -133,6 +133,8 @@ val pp_hist: Format.formatter -> histogram -> unit
 val tukey_fences : ?k:float -> float array -> float * float
 (** Refer to :doc:`owl_stats`. *)
 
+val iqr : float array -> float
```
ryanrhymes: Because we have used `first_quartile` and `third_quartile` as functions names, it makes more sense to use full name (i.e. `interquartile`) imo.

ryanrhymes:
I think `kde` should be part of stats module (in the base lib if it does not have any dependency on C code).  The implementation code looks good to me, let's see if @superbobry has further suggestions or comments?

Another thought, would polymorphic variants help in making the type definition lighter? But I am totally fine with the current type definition.

struktured:
@ryanrhymes factored the kde implementation as a `KDE` submodule of owl_base_stats, changed the variants to polymorphic versions, and renamed `iqr` to `interquartile`.

Let me know if you need any more tweaks before it's merge ready.

PR #304, opened 2018-09-11
Implementation of Lambda Array neuron in neural module

pvdhove:
More generalised version of the `Lambda` layer, with an arbitrary output shape and a function lambda that takes an array of node as an argument (instead of a single node).

Signed-off-by: pvdhove <pierre.vdhove@gmail.com>


ryanrhymes:
Looks very good to me, thank you Pierre! First PR to Owl 🎉

Issue #305, opened 2018-09-12
Time complexity of top and bottom functions for Ndarray


pvdhove:
The time complexity of [`_search_close_to_extreme`](https://github.com/owlbarn/owl/blob/master/src/owl/dense/owl_dense_ndarray_generic.ml#L5331), which is used by functions `top` and `bottom`, is currently O(n * k), where n is the number of elements in the ndarray and k is the number of extreme elements to find. However, a time complexity of O(n * log(k)) is achievable by replacing the array `vls` by a min/max heap of bounded size k (while keeping the same space complexity).

I don't think that there is a standard heap implementation in OCaml - would it make sense to implement a heap in Owl (in a file `base/misc/owl_utils_heap.ml` for instance)? If not, how would you structure this addition?

ryanrhymes:
I agree, the current `_search_close_to_extreme` has a very naive implementation. What you propose sounds very good so please go for that 👍

`base/misc/owl_utils_heap.ml` is the right place to implement the heap functionality, similar to `owl_utils_array.ml`.

PR #306, opened 2018-09-13
Implementation of heap + improved top and bottom + tests

pvdhove:
Fixes Issue #305.
Implementation of a min-heap using an array representation.
Use of this min-heap to improve the time complexity of functions `top x k` and `bottom x k` in the Ndarray module, from O(n * k) to O(n * log(k)).
The time taken by these functions is now much more predictable.
The worst case for `top` is when the ndarray is initially sorted in increasing order. This improvement can be tested with the following example:
```
let arr = Dense.Ndarray.D.init [|1000000|] (fun i -> float i) in
let result = Dense.Ndarray.D.top arr 100000 in
...
```
This used to take several minutes on my laptop, but it now takes only a few seconds.

### DIFF
--- a/src/base/misc/owl_utils.ml
+++ b/src/base/misc/owl_utils.ml
@@ -11,6 +11,8 @@ include Owl_utils_ndarray
 
 module Stack = Owl_utils_stack
 
+module Heap = Owl_utils_heap
+
 module Array = Owl_utils_array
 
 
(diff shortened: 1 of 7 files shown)


ryanrhymes:
Great PR! The code looks very neat and the improvement is significant. Many thanks Pierre!

ryanrhymes:
@pvdhove the documentation is also great, it has been rebuilt here: http://ocaml.xyz/apidoc/owl_utils_heap.html

PR #308, opened 2018-09-15
Neural module: add multi-input/output support

pvdhove:
I tried to minimise the changes to the current interface. I changed the `network` attribute `root : node option` to `roots : node array` (the case `None` is now equivalent to an empty array) and I added an attribute `outputs : node array`.
It should be good to always keep both `input`/`inputs`, `get_network`/`outputs` and `model`/`model_inputs`. However, in the future I think that `input_shape` and `run` can be replaced by `input_shapes` and `run_inputs`.
This adds more flexibility to the networks - it is good to rely on the `outputs` attribute rather than assuming that the desired output is the last node of `topo`.
A real multi-output training (as in other frameworks) is not supported yet. This will require some kind of `train_multi` function with arrays of nodes instead of nodes as arguments. This is significantly more difficult, so let me first know if you agree with those changes :). At the moment, multi-output networks can still be trained like normal networks (if the desired output is the last node of the topological order).
The new interface works on Mask R-CNN but it would be nice to find a smaller example to add it to the examples.

### DIFF
--- a/src/base/neural/owl_neural_graph.ml
+++ b/src/base/neural/owl_neural_graph.ml
@@ -34,21 +34,22 @@ module Make
     mutable train   : bool;       (* specify if a node is only for training *)
   }
   and network = {
-    mutable nnid : string;        (* name of the graph network *)
-    mutable size : int;           (* size of the graph network *)
-    mutable root : node option;   (* root of the graph network, i.e. input *)
-    mutable topo : node array;    (* nodes sorted in topological order *)
+    mutable nnid    : string;      (* name of the graph network *)
+    mutable size    : int;         (* size of the graph network *)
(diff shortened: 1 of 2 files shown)


ryanrhymes:
The modification looks very neat!  I agree with these changes. We can leave the multi-output training in future, your MRCNN task is more important.

Regarding the interface design, the current neuron creation functions (e.g. `linear`) in owl_neural_graph.ml takes one neuron as input then outputs one neuron (so that we chain them up to form a network using `|>`). Maybe we can consider to change the interfaces to take neuron array as input and output neuron array, so that we simplify the interface (e.g. unify `model` and `model_input`) but still are able to chain them up as before. 

Yes, an example would be very nice!

PR #309, opened 2018-09-19
Simplify modules aliases

teresy:
Build passes


teresy:
To be clear, this is just a syntactic change because `module X = struct include Y end` can be simplified to `module X = Y`. 

I.e., the `struct` ... `end` is redundant when doing an `include` on a single module.

PR #311, opened 2018-09-21
Add Delay operation to CG and optimised inference to neural_compiler.ml

pvdhove:
This PR adds
- `Delay`/`DelayArray` operations to the Computation Graph, to delay the evaluation of a function that is not tracked by CG. It would work to just call them `Lambda`/`LambdaArray` - however, in case the argument of the `Lambda` layer contains only supported operations, it would hide that function from CG (for example, the `lazy_mnist.ml` example would be slightly less efficient), so I think it is good to keep both kinds of operations.
- `model`/`model_inputs` functions in `owl_neural_compiler.ml` to provide an interface to optimise inference with CG as well. Note: something I don't like is that after copying the network (just like the standard `model` function in `owl_neural_graph.ml`), `Copy` nodes are created in the graph for every constant values. I am unsure if this is necessary and if there is a clean way around it.
- a test function to the `lazy_mnist.ml` with this new optimised inference to be more similar to `mnist_cnn.ml`. Optimised inference is not super useful for such a small example but for Mask R-CNN, it decreased the memory consumption from 11 Go to 7 Go.

### DIFF
--- a/examples/lazy_mnist.ml
+++ b/examples/lazy_mnist.ml
@@ -21,20 +21,40 @@ let make_network input_shape =
   |> linear 10 ~act_typ:Activation.(Softmax 1)
   |> get_network ~name:"mnist"
 
+let pack x = CGCompiler.Engine.pack_arr x |> Algodiff.pack_arr
+let unpack x = Algodiff.unpack_arr x |> CGCompiler.Engine.unpack_arr
 
 let train network =
   let x, _, y = Dataset.load_mnist_train_data_arr () in
-  let x = CGCompiler.Engine.pack_arr x |> Algodiff.pack_arr in
-  let y = CGCompiler.Engine.pack_arr y |> Algodiff.pack_arr in
+  let x = pack x in
+  let y = pack y in
   let params = Params.config
     ~batch:(Batch.Mini 100) ~learning_rate:(Learning_Rate.Adagrad 0.005) 0.1
     (* ~momentum:(Momentum.Standard 0.1) *)
   in
   CGCompiler.train ~params network x y
 
+let test network =
+  let imgs, _, labels = Dataset.load_mnist_test_data () in
+  let m = Dense.Matrix.S.row_num imgs in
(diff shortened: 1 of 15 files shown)


pvdhove:
Thank you for your comments @jzstark!

- My intention with Delay is to provide a way for the user to easily perform any function (not necessarily supported by CG) as a vertex in the computation graph. There were many functions that I found really hard to add to CG while implementing MRCNN, especially if they need non-ndarray values. For example, I had to gather elements from multiple ndarrays using the same array of n-dimensional indices. This requires integer arrays and I would have to split my function into multiple functions that are currently non-supported if I want to do it with CG. But with `Delay` I can just define my own function `f` to do that on `Dense.Ndarray.S` (instead of `Owl_computation_cpu_engine.Make (Dense.Ndarray.S)`) and then call `(delay f)` instead of `f`.
To use `Delay` with the `Lambda` layer, if `f` is not fully supported by CG but is supported by `Algodiff`, you have to do `lambda (delay f) node`.
The limitation about the shape can indeed be removed by adding a `shape` parameter (just like `DelayArray`). This is more verbose and if the shape needs to be changed, `DelayArray` can be used. I also wanted it to be similar to the behaviour of `Lambda`. Of course, @ryanrhymes let me know if you are OK with this design and with the names of the functions. I'll try to think about a short example in which `Delay` is absolutely necessary (whether it is in a DNN or not).
- The two `model` functions have the same behaviour: it works to call `model nn input` to get the result. In both cases, `let eval = model nn in ...` performs a pre-computation on the network, so that `eval` can be called on multiple different inputs without re-copying the network every single time. Maybe the design can be changed to enforce that behaviour? It might not be obvious to the user that he shouldn't call `model network input` multiple times with the same `network`.
- In `owl_computation_cpu_init.ml`, that's indeed what I meant by 'imperative'. It is possible with a while loop and `ref`s but in OCaml that would be uglier, I'll just remove the comment.

I'll fix the comments soon, thanks for noticing that. I forgot the call to `_remove_training_nodes` as well to optimise `model` more.

PR #312, opened 2018-09-26
Add verbosity flag to kmeans function

chrismamo1:
it can be annoying to clog my terminal with log messages when I'm clustering in a deeply nested loop

I'm not sure if this is totally consistent with the rest of Owl's source, but it's a minor change so I can't see how I could have messed it up that badly.


pvdhove:
You could also change the log level in your code with `Owl_log.set_level WARN;` to disable the `INFO` logs.

Issue #288, opened 2018-08-21
Import non-weight values of Normalisation neuron


jzstark:
In neural network, a Normalisation neuron contains `mu` and `var` parameters, besides `gamma` and `beta`. Its behaviour varies in different mode: it normalises input with average and variance of current batch in training mode and keep updating the moving `mu` and `var`;  in inference mode it normalises input directly with the calculated `mu` and `var` from training phase. Unlike `gamma` and `beta`, these two are not weight values.

One implication of this difference is that, when we are importing values from existing pre-trained models, we cannot directly use `Neuron.update` to import the value of `mu` and `var` into a network. My current solution is to manually update these values in the normalisation neurons in a network, and then save the network itself instead of its weight.  As a result, for image classification networks such as Inception and ResNet, by saving the network, we cannot then change its input sizes. 

So I'm looking at either a better alternative to this current practice in importing values from existing models (e.g. maybe using the OpenVX/NNEF format standards Liang has mentioned), or a mechanism to import non-weight trainable values following that of existing frameworks such as Keras/MxNet/etc. 

In Keras, each weight variable is assigned "trainable" or "non-trainable" -- the latter only consists of `mean` and `variance` in the BatchNormalization layer.  The weight loading/saving does not distinguish between these two kinds of weights.

pvdhove:
Here is what I could find in other frameworks:
- in Keras, each layer has two list attributes `trainable_weights` and `non_trainable_weights`. The function `get_weights` simply returns the concatenation of the lists. Weights are wrapped in a `Variable` (similar to a `Var` node from Owl's CG) and must be added to a layer with a function `add_weights(shape, name, initializer, regularizer, trainable)`. This function returns the `Variable` object and adds it to either `trainable_weights` or `non_trainable_weights`. Since the value of the weights is wrapped into a `Variable`, both the weight and its reference in the list can point to the same ndarray value (in Owl, the weight is the ndarray value so that is not currently possible).
- in Tensorflow, weights are `variable`s and their values are saved when you save a model. The training part is a bit uglier.
- PyTorch distinguishes parameters (W, b) from 'buffers' (mean, var). Buffers are only explicitly used for the normalisation neuron. The `save` function saves both parameters and buffers. BatchNorm also overwrites the `load_from_state_dict` function and is the only neuron doing so.

[...]

PR #317, opened 2018-10-18
Save/load normalisation neuron

pvdhove:
Fixes issue #288.


ryanrhymes:
Neat! Thanks for fixing it 💯

PR #318, opened 2018-10-22
New memory optimisation algorithm for CG (+ block attribute, multimap)

pvdhove:
This PR consists of:
- A Multimap implementation in `Owl_utils` (extension of OCaml's `Map` which allows the same key to be bound to multiple values) and one small test file for this structure (+ a small fix of `owl_utils_stack.ml`).
- A change of the `vnode` parameter of CG's nodes to a more general type `block`. A `block` holds the reference to an ndarray of fixed size. Multiple nodes can share the same block.
- A new algorithm to initialise the memory of the computation graph (inspired by [the one used by MxNet](https://mxnet.incubator.apache.org/architecture/note_memory.html#memory-allocation-algorithm)). The main idea is that during a graph traversal in evaluation order, we can spot a node which doesn't need its value anymore by keeping a counter of how many children still have to use that node. The block of memory used by that node can thus be tagged as reusable. When allocating the memory to a node, we then know which blocks can be reused. To select a block to allocate to a node, the following heuristic is used: the smallest available block that is larger than the size of the node is chosen. If such a block does not exist, the largest block is chosen and its size is increased. A multimap is needed to do this efficiently.
We can note that the allocation strategy depends on the order of the traversal, and the evaluation **must** be done in the same order as the initialisation. This implementation uses a DFS order (shorter to implement and BFS was not necessarily better).
The time complexity is `O(n * log(b))`, where `n` is the number of nodes of the computation graph and `b` is the number of blocks allocated at the end of the algorithm (in theory, `b <= n` but in practice, `b << n`). The space complexity is `O(n + b)`.

You can find **[here](https://drive.google.com/open?id=12KCY9OC6GjuHiH2pRiAjqNi-pz2sNcc1)** some visual examples of the allocation performed by the algorithm (I recommend opening the big `.svg` files with Chrome or Firefox). Each reusable node is coloured depending on the block used (if the colours are too hard to distinguish, the `b` attribute is the id of the block). InceptionV3 needs only 6 shared blocks and the inference for `lazy_mnist` only 3.

### DIFF
--- a/src/base/compute/owl_computation_cpu_engine.ml
+++ b/src/base/compute/owl_computation_cpu_engine.ml
@@ -25,8 +25,9 @@ module Make_Nested
 
   (* core interface *)
 
+  (* ! Make sure the order of evaluation of init and eval is always the same. *)
   let eval_gen nodes =
-    Array.iter CG_Init._init_term nodes;
+    CG_Init._init_terms nodes;
     Array.iter CG_Eval._eval_term nodes
 
 
--- a/src/base/compute/owl_computation_cpu_engine.ml
+++ b/src/base/compute/owl_computation_cpu_engine.ml
@@ -28,7 +28,7 @@ module Make_Nested
   (* ! Make sure the order of evaluation of init and eval is always the same. *)
   let eval_gen nodes =
     CG_Init._init_terms nodes;
-    Array.iter CG_Eval._eval_term nodes
+    CG_Eval._eval_terms nodes
 
 
   let eval_elt xs = Array.map elt_to_node xs |> eval_gen
--- a/src/base/compute/owl_computation_cpu_eval.ml
+++ b/src/base/compute/owl_computation_cpu_eval.ml
@@ -19,11 +19,7 @@ module Make
 
   (* utility functions *)
 
-  let invalidate_opt x_option =
-    match x_option with
-    | Some x -> invalidate x
-    | None   -> ()
-
+  let invalidate_opt = function Some x -> invalidate x | None -> ()
 
   let update_validity x b =
     invalidate_opt (get_active_node b);
--- a/src/base/compute/owl_computation_cpu_eval.ml
+++ b/src/base/compute/owl_computation_cpu_eval.ml
@@ -33,214 +33,216 @@ module Make
 
   (* core evaluation function *)
 
-  let rec _eval_term x =
-    Owl_log.debug "eval %s ..." (node_to_str x);
+  let rec _eval_terms nodes =
+    let eval x =
+      Owl_log.debug "eval %s ..." (node_to_str x);
 
-    if is_valid x = false then
-      let () = try
-        match get_operator x with
-        | Noop                                          -> _eval_map_00 x (fun x -> x.(0))
-        | Var                                           -> check_assigned x
-        | Const                                         -> check_assigned x
-        | Empty _shape                                  -> _eval_map_01 x (fun ~out _x -> ()) [@warning "-27"]
-        | Zeros _shape                                  -> _eval_map_01 x (fun ~out _x -> A.zeros_ ~out)
-        | Ones _shape                                   -> _eval_map_01 x (fun ~out _x -> A.ones_ ~out)
-        | Create _shape                                 -> _eval_map_02 x (fun ~out x -> A.create_ ~out x.(0))
-        | Sequential _shape                             -> _eval_map_02 x (fun ~out x -> A.sequential_ ~a:x.(0) ~step:x.(1) ~out)
-        | Uniform _shape                                -> _eval_map_02 x (fun ~out x -> A.uniform_ ~a:x.(0) ~b:x.(1) ~out)
-        | Gaussian _shape                               -> _eval_map_02 x (fun ~out x -> A.gaussian_ ~mu:x.(0) ~sigma:x.(1) ~out)
-        | Bernoulli _shape                              -> _eval_map_02 x (fun ~out x -> A.bernoulli_ ~p:x.(0) ~out)
-        | Init (_shape, _f)                             -> failwith "Init"
-        | Get i                                         -> _eval_map_06 x (fun x -> A.get x i)
-        | Set _i                                        -> failwith "Set"
-        | GetSlice slice                                -> _eval_map_01 x (fun ~out x -> A.get_slice_ ~out slice x.(0))
-        | SetSlice slice                                -> _eval_map_01 x (fun ~out x -> A.set_slice_ ~out slice x.(0) x.(1))
-        | Copy                                          -> _eval_map_01 x (fun ~out x -> A.copy_ ~out x.(0)) [@warning "-27"]
-        | Reset                                         -> _eval_map_01 x (fun ~out _x -> A.reset out)
-        | Reshape _shape                                -> _eval_map_01 x (fun ~out x -> A.reshape_ ~out x.(0))
-        | Reverse                                       -> _eval_map_01 x (fun ~out x -> A.reverse_ ~out x.(0))
-        | Tile repeats                                  -> _eval_map_01 x (fun ~out x -> A.tile_ ~out x.(0) repeats)
-        | Repeat repeats                                -> _eval_map_01 x (fun ~out x -> A.repeat_ ~out x.(0) repeats)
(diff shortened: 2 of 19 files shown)


src/base/types/owl_types_computation_device.ml
```diff
@@ -14,7 +14,7 @@ module type Sig = sig
   type device
   (** TODO *)
 
-  type value
+  type value = ArrVal of A.arr | EltVal of A.elt
```
ryanrhymes: Can we keep this type definition abstract? This is device dependent, if we expose the type definition here, it becomes difficult to implement CG on other devices. 

Based on the two pattern matchings you used, is it possible to use `owl_computation_symbol.is_elt` instead to achieve the same thing?

ryanrhymes: There are also `owl_computation_symbol.is_arr` and `owl_computation_symbol.node_numel` functions, might be helpful?

src/base/compute/owl_computation_cpu_engine.ml
```diff
@@ -25,8 +25,9 @@ module Make_Nested
 
   (* core interface *)
 
+  (* ! Make sure the order of evaluation of init and eval is always the same. *)
   let eval_gen nodes =
-    Array.iter CG_Init._init_term nodes;
+    CG_Init._init_terms nodes;
     Array.iter CG_Eval._eval_term nodes
```
ryanrhymes: A suggestion, given `CG_Init._init_terms nodes;` I think we can also add `CG_Eval._eval_terms nodes;`, for being consistent and better readability.

src/base/types/owl_types_computation_device.ml
pvdhove: Actually using `owl_computation_symbol.is_elt` in `make_value_block` breaks the `unit_lazy.ml` tests, because it relies on the shape of the node when `make_node` is called ─ in `unit_lazy.ml`, I think the shape is never computed. However would it be reasonable to add an `is_elt : value -> bool` and `is_arr : value -> bool` to `owl_types_computation_device.ml`? The current interface already mentions `elt`'s and `arr`'s and it might be cleaner than using the shape to distinguish them? :)

src/base/compute/owl_computation_cpu_engine.ml
pvdhove: That's right, I'll fix it.

src/base/types/owl_types_computation_device.ml
ryanrhymes: Yes, let's move them to `owl_types_computation_device.ml`. Because `value` is device-dependent, so it should be device's job to tell if it is a scalar or ndarray. Logically, it makes more sense, I agree.

src/base/compute/owl_computation_cpu_init.ml
```diff
@@ -14,321 +14,296 @@ module Make
...
+    | EltGreater      -> true
+    | EltLessEqual    -> true
+    | EltGreaterEqual -> true
+    | _               -> false
+
+
+  (* written to be as safe as possible, but can probably set to true more
+   * operations. *)
+  let can_overwrite_parents x =
+    match get_operator x with
```
ryanrhymes: as far as I can tell, many operations such as pooling and conv can be set to `true`. You can try to confirm with @jzstark 

Previously, I used init_00 which allocates the new memory because the input and output shape do not match, since your new algorithm can deal with this, so these can be set to true imo, to save even more memory.

ryanrhymes:
> * Replaced `_eval_term` with `_eval_terms`.
> * Added `is_elt` and `is_arr` in `owl_computation_device`.
> * I replaced `owl_computation_symbol.is_elt`/`arr` by `is_node_elt`/`arr` to avoid the ambuigity and simply renamed the calls to these functions. It might be interesting to have a slightly cleaner way to distinguish node elements and node arrays (rather than relying on their `shape` or now `value` which are not always available) but probably not in this PR.

yes, i agree. The interface needs some cleaning up ... sometimes even i am a bit confused. but i guess value is much more reliable than shape, isn't it?

src/base/compute/owl_computation_cpu_init.ml
pvdhove: Yes, I based most of the current classification on the previous one with `init_00`. However I think `conv` cannot overwrite its parents while it is being computed, since the result of one cell depends on the value of the adjacent cells ─ so if those were modified during the computation, the result would not be correct. `pool` could be fine but it depends in which order the computation is done. Since I don't want to rely on implementation details, I prefer leaving it `false`. The person who implemented it (@jzstark) might confirm that it can overwrite the parents though :)

pvdhove:
Yes, here I used `is_elt` and `is_arr` for actual `value` types (not nodes) so it's fine. However, checking `is_elt (attr x).value` to check if a node `x` is an element is not necessarily more reliable than the shape since usually the value of the node is computed later than the shape.

ryanrhymes:
> Yes, here I used `is_elt` and `is_arr` for actual `value` types (not nodes) so it's fine. However, checking `is_elt (attr x).value` to check if a node `x` is an element is not necessarily more reliable since usually the value of the node is computed later than the shape.

maybe `op` type in `owl_computation_type` is actually a better option in the end, w.r.t deciding if the output is scalar or array ...

pvdhove:
> maybe `op` type in `owl_computation_type` is actually a better option in the end, w.r.t deciding if the output is scalar or array ...

Yes, you are right. Let me know if you'd like to have this change in the same PR, it should not be too difficult to implement.

ryanrhymes:
> > maybe `op` type in `owl_computation_type` is actually a better option in the end, w.r.t deciding if the output is scalar or array ...
> 
> Yes, you are right. Let me know if you'd like to have this change in the same PR, it should not be too difficult to implement.

Let's leave it for another separate PR, before you write up your report to present CG and related optimisation stuff as a whole. Then we need to clean up the api a bit. For the time being, let's focus on memory opt which itself is already a big PR.

src/base/compute/owl_computation_cpu_init.ml
ryanrhymes: agree. @jzstark took a similar strategy while optimising underlying tensor ops, starting from general optimisation, then boils down to case by case optimisation.

src/base/compute/owl_computation_cpu_eval.ml
```diff
@@ -19,232 +19,241 @@ module Make
 
   (* utility functions *)
 
-  let update_validity x =
-    validate x;
-    Array.iter invalidate (get_vnode x)
+  let invalidate_opt x_option =
+    match x_option with
+    | Some x -> invalidate x
+    | None   -> ()
```
ryanrhymes: `let invalidate_opt = function Some x -> invalidate x | None -> ()`, just a shorter version if you need.

src/base/compute/owl_computation_symbol.ml
```diff
@@ -376,32 +465,30 @@ module Make
   let is_const x = (attr x).op = Const
 
 
-  let is_arr x =
+  let is_node_arr x =
```
ryanrhymes: let's leave a mark here to remind us to clean up the interface later. e.g., using `op` type for a more reliable version of `is_elt` and `is_arr`

src/base/compute/owl_computation_symbol_sig.ml
```diff
@@ -89,35 +125,41 @@ module type Sig = sig
   val set_reuse : attr Owl_graph.node -> bool -> unit
   (** TODO *)
 
-  val get_reuse : attr Owl_graph.node -> bool
-  (** TODO *)
-
-  val set_vnode : attr Owl_graph.node -> t array -> unit
+  val is_reusable : attr Owl_graph.node -> bool
```
ryanrhymes: Can we change this back to `get_reuse` if you dont have strong objection? Just for being consistent with other getter and setter functions.

pvdhove:
> > Yes, here I used `is_elt` and `is_arr` for actual `value` types (not nodes) so it's fine. However, checking `is_elt (attr x).value` to check if a node `x` is an element is not necessarily more reliable since usually the value of the node is computed later than the shape.
> 
> maybe `op` type in `owl_computation_type` is actually a better option in the end, w.r.t deciding if the output is scalar or array ...

About this: a problem is that `Const` and `Var` nodes can be either scalars or arrays, so it is not possible to solely use `op` to check whether a node is scalar or not. `Var` nodes might not even be initialised, which means that when they are created there is no way to know if they are scalar or not.

PR #320, opened 2018-11-01
Switch to dune.configurator

diml:
The `configurator` package has been replaced by the `dune.configurator` library. This PR changes owl to use the latter.

### DIFF
--- a/owl.opam
+++ b/owl.opam
@@ -21,7 +21,6 @@ depends: [
   "base" {build}
   "base-bigarray"
   "conf-openblas" {>= "0.2.0"}
-  "configurator" {build}
   "ctypes"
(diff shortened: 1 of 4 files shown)


ryanrhymes:
Thank you @diml for fixing this 👍

ryanrhymes:
this is due to the inconsistent interface of caml used by dune and and base (note not owl's base lib). i have replaced base with sexplib, submitted the fix in https://github.com/owlbarn/owl/commit/3615105a9a79980add9e9682381c495e2b1454f9

it should work now, give it a try and let me know if there are still issues.

diml:
This should also work with recent versions of Dune (1.4 IIRC).

PR #324, opened 2018-11-09
Finish the is_prime function

ejaasaari:
Finishes the incomplete is_prime function. The algorithm used is the Miller-Rabin primality test with bases that make it deterministic for all integers that an OCaml int can represent.


ryanrhymes:
Thanks, I will also rebuild the doc on ocaml.xyz later.

PR #325, opened 2018-11-09
Print node in CG module

pvdhove:
Printing the outputs of intermediate nodes in a computation graph is not possible because there is currently no way to make the print lazy, which makes debugging quite hard. I've added a `LazyPrint` node which is an identity node having the side effect of printing the output of its parent when evaluated (same trick as [TensorFlow Print()](https://www.tensorflow.org/versions/r1.11/api_docs/python/tf/Print)).

The way to use it is to insert it after the node whose output you want to print, for instance
```
module Engine = Owl_computation_cpu_engine.Make(Dense.Ndarray.S)

let x =
  Engine.gaussian [|100; 100|]
  |> Engine.sqr
  |> Engine.lazy_print ~max_row:5 ~max_col:5
  |> Engine.floor
in
Engine.eval_arr [|x|]
```

You can also use it in a neural network using CG quite easily, by putting it in a Lambda layer:
```
  |> lambda (fun x -> unpack_arr x
                      |> Engine.lazy_print ~max_row:5 ~max_col:5
                      |> pack_arr)
```
where unpacking the CG node and standard printing would not work because the node is not evaluated when the graph is built.

Note that it needs to be a different function from `print` because it has to return an `arr` to work, but `print` has to return `unit`.

### DIFF
--- a/src/base/compute/owl_computation_cpu_eval.ml
+++ b/src/base/compute/owl_computation_cpu_eval.ml
@@ -69,6 +69,7 @@ module Make
           | OneHot depth                                  -> _eval_map_01 x (fun ~out x -> A.one_hot_ ~out depth x.(0))
           | Delay f                                       -> _eval_map_08 x f
           | DelayArray (_shape, f)                        -> _eval_map_00 x f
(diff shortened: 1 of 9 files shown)


ryanrhymes:
A great addition to the current CG functionality, thank you Pierre 👍

PR #327, opened 2018-11-11
Clean split_parents function in CG cpu_init

pvdhove:
Modify the `split_parents` function in `owl_computation_cpu_init.ml` to make it more similar to the other CG files, to be easier to optimise and to generate a compilation error if a new operation is added without changing this file.

Fix the status of `Reverse` and the `TransposeConvNd` operations (cannot overwrite their parents) + very small optimisation for `Concatenate`. I have added a comment next to the operations that might maybe overwrite some of their parents depending on their parameters, in case the person who implemented them wants to think about it :)

### DIFF
--- a/src/base/compute/owl_computation_cpu_init.ml
+++ b/src/base/compute/owl_computation_cpu_init.ml
@@ -25,109 +25,226 @@ module Make
 
   (* utility functions *)
 
-  let is_broadcastable x =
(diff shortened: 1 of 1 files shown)


ryanrhymes:
Yes, new interface indeed looks much more consistent with other modules. Thank you Pierre 👍

PR #330, opened 2018-11-16
Fix cache size check

jzstark:
This PR adds support for more CPU infrastructures when checking L1/L2/L3 cache sizes, which affect the convolution computation.


ryanrhymes:
thank you Jianxin. Can we possibly move these functions of detecting CPU and L1/2 cache size to `owl_core_utils.c`? They may be useful to other functions as well in future.

PR #332, opened 2018-11-20
Add initial AEOS module to Owl

jzstark:
This PR contains an initial framework of AEOS (Automated Empirical Optimisation of Software) module in Owl, with OpenMP threshold tuning as examples in this framework. 

This module tries to minimise both dependency and changes of Owl library itself. The major change at core modules is that, whereas previously we use only one threshold value, now we need different threshold values, so extra macro definition is added.  The OpenMP version of `sum` and `prod` operation is also added. 

Current supported operations in AEOS module is unary and binary map operations, with a small number of fold operations. Each operation to be turned is one stand-alone module, so that possible future extension should be straightforward. 

This PR is still an on-going work and, though after several checks, may contains issues from system structure to code format. I'll try to finishing polishing it within this week,  and any feedback will be quite appreciated.


ejaasaari:
A small suggestion: I previously did similar-ish work and found the [Theil-Sen estimator](https://en.wikipedia.org/wiki/Theil%E2%80%93Sen_estimator) to be better for modeling timings robustly than trying to remove outliers in simple linear regression (and it's equally easy to implement) :)

jzstark:
Thanks a lot for this information @ejaasaari ! It indeed looks like a better approach to be used here. I'll use it to enhance the current regression step.

src/owl/config/configure.ml
```diff
@@ -82,7 +83,7 @@ let get_gcc_path _c =
 
 
 let get_openblas_path _c =
-  let p0 = "/usr/local/opt/openblas/lib" in
+  let p0 = "/usr/local/lib/openblas/lib" in
   let p1 = "/opt/OpenBLAS/lib/" in
```
ryanrhymes: configure.ml require extra attention whenever you change it, because something works on your system does not necessarily mean it will work on others. your previous change definitely will break it on some system, e.g. mine. however, you can extend it by adding more rules to detect.

ryanrhymes:
Great PR!

Issue #335, opened 2018-12-04
gcc doesn't recognize -framework option


emengd:
Hi again,

Trying to compile the current master branch, I get the following error :

```
dune external-lib-deps --missing @install @runtest
dune build @install
  ocamlmklib src/aeos/dllowl_aeos_stubs.so,src/aeos/libowl_aeos_stubs.a (exit 2)
(cd _build/default && /home/emile/.opam/base/bin/ocamlmklib.opt -g -o src/aeos/owl_aeos_stubs src/aeos/owl_aeos_tuner_map_stub.o src/aeos/owl_aeos_tuner_fold_stub.o -lm -lomp -framework Accelerate)
gcc: error: Accelerate: No such file or directory
gcc: error: unrecognized command line option ‘-framework’
Makefile:13: recipe for target 'build' failed
make: *** [build] Error 1
```

Removing the offending option from `src/aeos/dune` let me compile.

I use Linux Mint 18 (Ubuntu 16.04) with gcc 5.4.0.

jzstark:
Hi @emengd, thanks for finding out this issue. The aeos module is still in development and currently contains several compile options that are not cross-platform as you've found. One solution could be to include a configuration module, such as [src/owl/config](https://github.com/owlbarn/owl/tree/master/src/owl/config) to choose options depending on different platform.

PR #336, opened 2018-12-05
Add configuration file for aeos module and update Dockerfiles

jzstark:
This PR adds a configuration file trying to address #335. This config file is similar to [src/owl/config](https://github.com/owlbarn/owl/tree/master/src/owl/config), with smaller number of compiling options. 

I've tested compiling Owl on MacOS, RaspberryPi, Ubuntu, and several docker containers. Current Dockerfiles are therefore also fixed in this PR.


ryanrhymes:
great, thanks for fixing this, jianxin 👍

PR #343, opened 2019-01-14
unit test for reverse-mode matrix operations and new functions in algodiff

tachukao:
1. Added a new unit test for reverse-mode matrix operations `test/unit_algodiff_matrix.ml`. This test is based on comparing the algodiff gradients with gradients estimated using finite difference.

2. Implemented several new functions in reverse-mode differentiation: `trace`, `diag`, `tril`, `triu`, `qr`, `lyapunov`


ryanrhymes:
Awesome! Thank you both Guillaume and Calvin! 🥇

PR #345, opened 2019-01-16
simpler way to deal with multi-output functions in reverse-mode + expose gradient test to user

tachukao:
1. modified type `DR` in algodiff to track when to update adjval of inputs in reverse-mode e.g. `Algodiff.D.Maths.qr` now has type `t ->  t*t`
2. `FDGrad_test`: a new functor in algodiff that takes a module of user-specified parameters and can be used to test gradients (i.e. compare reverse-mode gradients with finite difference approximations)


ryanrhymes:
Thank you Calvin, awesome!

Issue #341, opened 2019-01-06
Optimise consecutive calls to Stats.quantile


pveber:
The [current implementation](https://github.com/owlbarn/owl/blob/master/src/base/stats/owl_base_stats.ml#L554) is not efficient when one wants to compute several quantiles at a time, because it will sort the sample at each call. What about the following simple change:
```ocaml
let quantile x =
  let y = sort ~inc:true x in
  let n = Array.length y in
  fun p ->
    let index = p *. [...]
```
with a corresponding explanation in the doc?

BTW, there used to be a check on `p` (should be between 0 and 1), one could raise `Invalid_argument` if not.
I can send a PR if that helps.

ryanrhymes:
The proposal looks good to me, please do submit a PR. many thanks!

my apologies for the delay in replying :)

pveber:
No worries, here it is!

PR #346, opened 2019-01-17
stats: optimisation of quantile for repeated calls (#341)

pveber:
Here's a proposition for #341, I'm not too sure about the syntax for documentation, it seems it is different from the usual `ocamldoc` syntax, right?

### DIFF
--- a/src/base/stats/owl_base_stats.ml
+++ b/src/base/stats/owl_base_stats.ml
@@ -551,18 +551,22 @@ let pp_hist formatter hist =
 
 
 
-let quantile x p =
+let quantile x =
   let y = sort ~inc:true x in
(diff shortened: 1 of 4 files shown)


pveber:
it seems only `opam lint` is failing here.

pveber:
Added a single one against `R`. Since there are two implementations (one in OCaml, the other in C), would you like me to add another to test the OCaml version?

ryanrhymes:
> Added a single one against `R`. Since there are two implementations (one in OCaml, the other in C), would you like me to add another to test the OCaml version?

Yes please do, many thanks. I will merge current PR, it looks very good to me already, good job 👍  You can submit the unit test for ocaml part in another PR?

PR #360, opened 2019-02-01
owl: add toplevel_printer annotation

mseri:
This will make the pretty printers automatically installed
in recent versions of utop (and hopefully further toplevels,
like jupyter, will follow). Practically it means that owl-top is
no longer necessary for utop to pretty print owl values.
See e.g. https://github.com/ocaml/opam-repository/pull/13348

### DIFF
--- a/src/base/core/owl_graph.mli
+++ b/src/base/core/owl_graph.mli
@@ -184,7 +184,7 @@ Topological sort of a given graph using a DFS order. Assumes that the graph is
 acyclic.
  *)
 
-val pp_node : Format.formatter -> 'a node -> unit
+val pp_node : Format.formatter -> 'a node -> unit [@@ocaml.toplevel_printer]
 (** Pretty print a given node. *)
 
 val to_string : bool -> 'a node array -> string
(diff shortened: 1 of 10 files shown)


ryanrhymes:
Awesome feature, let me know in case anything on [this page](http://ocaml.xyz/chapter/interact.html) needs to be revised accordingly, thanks!

mseri:
The only change is that since utop 2.3.0 and with this change, `#require "owl-top";;` is no longer required. I need to try with jupyter and in case open an issue there if it does not yet recognise it as well

dbuenzli:
> I need to try with jupyter and in case open an issue there if it does not yet recognise it as well

I guess this is unlikely since AFAIK `jupyter` is using OCaml's toploop library not `utop`. However do not open an issue with jupyter, the right place to fix this is upstream, see [this issue](https://caml.inria.fr/mantis/view.php?id=7770) if you are intersted in working on it (and ignore @diml last message not everyone agrees with his assessment).

mseri:
Thanks @dbuenzli I was not aware of that issue

diml:
I wasn't aware that there was a disagreement on the upstream issue. I added a message to clarify things.

PR #371, opened 2019-02-13
current version of eigen 0.0.6 on opam does not have `Eigen.Types`

tachukao:
use `Eigen_types` instead of `Eigen.Types` for compatibility


ryanrhymes:
Marcello recently upgrade eigen to 0.1.0, which introduced some breaking changes. Can you upgrade to eigen 0.1.0? @mseri comment?

tachukao:
Argh never mind, I didn't realise that my opam was outdated and couldn't find the latest version of eigen. After updating, everything works fine.

PR #372, opened 2019-02-13
base: remove Unix time dependency temporarily

ehirdoy:
Remove Unix time code from Owl_base.

This is needed because MirageOS failed to be built as:

 Error: No implementations provided for the following modules:
          Unix referenced from /home/ehirdoy/.opam/4.06.1+lto/lib/owl-base/owl_base.cmxa(Owl_utils),
            /home/ehirdoy/.opam/4.06.1+lto/lib/owl-base/owl_base.cmxa(Owl_log),
            /home/ehirdoy/.opam/4.06.1+lto/lib/owl-base/owl_base.cmxa(Owl_io)

This can be temporary workaround.
Other non-unix time function could revert this commit later.

### DIFF
--- a/src/base/misc/owl_io.ml
+++ b/src/base/misc/owl_io.ml
@@ -46,20 +46,13 @@ let iteri_lines_of_file ?(verbose=true) f fname =
   let i = ref 0 in
   let h = open_in fname in
   (
-    let t0 = Unix.gettimeofday () in
-    let t1 = ref (Unix.gettimeofday ()) in
     try while true do
       f !i (input_line h);
       i := !i + 1;
       (* output summary if in verbose mode *)
       if verbose = true then (
-        let t2 = Unix.gettimeofday () in
-        if t2 -. !t1 > 5. then (
-          t1 := t2;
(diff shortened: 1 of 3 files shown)


mseri:
Some of these could be easily swapped to use https://github.com/dbuenzli/mtime
I think that should work with mirage or am I wrong?

ehirdoy:
@mseri thanks for the tips!! I'll take a look later

ehirdoy:
Isn't it better to keep this in issue as ToDo?

ehirdoy:
For the record, from slack discussion,

Right, unknown reference to ocaml_[m,p]time_**clock**_?_ns. I think that [m,p]time itself is useless without [m,p]time_clock since clock is the one actually accessing to the clock source??? And [m,p]time.os doesn't have its backend for solo5 API. Instead, Mirage has Mirage_clock.PCLOCK with solo5 backend, but using this in Owl doesn't seem so nice :confused: but this seems the only solution to me for now.....Probably I'll survive with applying temporarily remove gettimeofday patch for my Mirage build until there will be better solution for this.....here's a branch if someone wants to dig in :wink: https://github.com/ehirdoy/owl/commits/gettimeofday

PR #380, opened 2019-02-18
Further fixups

mseri:
The depext is to make the build work again on freshly installed macs, the dune file change is just to make the update from the previous PR behave exactly as before (almost: I kept `safe-string` out as it is the default now and can create runtime and compiler error when libraries are compiled with mixed safe/unsafe strings in old compilers).

### DIFF
--- a/owl-aeos.opam
+++ b/owl-aeos.opam
@@ -20,3 +20,7 @@ depends: [
   "ocaml-compiler-libs"
   "stdio" {build}
 ]
+
+depexts: [
+  ["libopenmp"] {os = "macos" & os-distribution = "homebrew"}
+]
--- a/src/owl/dune
+++ b/src/owl/dune
@@ -165,7 +165,7 @@
   )
  (c_flags (:include c_flags.sexp))
  (c_library_flags (:include c_library_flags.sexp))
- (flags (:include ocaml_flags.sexp))
+ (flags :standard (:include ocaml_flags.sexp))
  (libraries
   ctypes
   ctypes.stubs


ryanrhymes:
Does this mean people have to install openmp in order to install owl? Do we really want to enforce this dependency (especially most users probably won't enable openmp)? By default, OpenMP is disabled and Owl can certainly run without it. @mseri @jzstark

mseri:
This was already the case but it breaks the build if your mac doesn’t have openmp (like mine)

ryanrhymes:
Ah, this is a bug then, because it should break the build if the user doesn't have OpenMP installed. @jzstark I think this is due to AEOS which is responsible for tuning OpenMP parameters, even when openmp is disabled, the build process still relies on AEOS to generate a dummy header file to continue. Jianxin, I think you probably need to fix this by having a dummy header file in the place by default, if OpenMP is enable, let AEOS overwrite the default header file to use the best parameters.

Also, if OpenMP is disabled, we should not compile AEOS module at all in the first place to avoid breaking the build process.

mseri:
Btw I have nothing baout disabling openmp, I nave added the depext because it was the way I made the master branch of owl work on my mac

jzstark:
Yes I see. The AEOS configuration is quite rudimentary. I've changed its dune file recently to make it from a library to executables. I'll take a look at removing its dependency on OpenMP when AEOS is disabled by the dummy file solution. However, since I've tested it on all kinds of containers and still failed to find this issue the last time, do you have any suggestion about how to disable the OpenMP for testing? @mseri

ryanrhymes:
@jzstark dummy header is the cleanest solution AFAIK, because owl needs that header file anyway which should NOT have dependency on AEOS, especially when the user does not want to enable OpenMP. In short, AEOS should not be compiled and run if openmp switch is off.

mseri:
I agree on the dummy header. We could simply make owl-aeos a depopt: we can use the select function from dune to pick the right header depending on owl-aeos being present or not.

mseri:
Closing this as we are discussion aeos fix somewhere else. 
I am moving the reintroduction of the `:standard` flags to a separate PR

PR #383, opened 2019-02-21
Modify aeos config and build files

jzstark:
This PR updates the configuration and build files of AEOS module. After some discussion with Liang, we think it's best to provide two dune files: a normal one that does not contain AEOS, and a `dune.omp` that can be manually renamed to `dune` if the user what to try AEOS with OpenMP. A dummy file `owl_aeos_params_generated.h` is used in the absence of the AEOS module. Please take a look at the code to see if it helps to solve your current issue @mseri . Also, the AEOS module does not build a public library now, for TravisCI test to pass.


mseri:
I think this is the wrong approach: there is no need to make it so cumbersome imho.
I think the simplest solution is to use the discover mechanism (see https://dune.readthedocs.io/en/latest/configurator.html#usage) to check if `libomp` is usable (and if you want a flag like owl already does) and produce one or the other header depending on the case. This same mechanism can be used to detect the need for `-llapacke` when linking.

Alternatively we could simply move `owl-aeos` into a `depopts` stanza in the owl opam file. Then make owl-aeos install the c header that it generates (with install_c_headers in the dune file), and use https://dune.readthedocs.io/en/latest/dune-files.html#alternative-dependencies inside the owl dune file to get the dummy headers from a local path if needed (or use the ones installed by aeos). In this case it would be enough to add the depexts on libomp to the owl-aeos opam file to fix the tests.

jzstark:
Thanks for the help @mseri ! I'll try to fix it as suggested. Amid the current change of Owl/Eigen/Travis etc. configuration, I think it would be better for me to do it later when things are settled down.  If you find AEOS/OpenMP configuration stands in your way, please do no hesitate to disable it for now.

ryanrhymes:
Can we split out AEOS by using a dummy header, at least for the time being. Because AEOS is an executable now, so someone can always just run it to get optimised openmp parameters. We can think later how to incorporate better and smarter method in conditional compilation. As least, AEOS will not block the normal building without openmp installed.

jzstark:
Agree. I think we can just make aeos an executable and move it out of the dune file for now, just to make current configuration matter a bit simpler. I've just sync this PR with master branch.

mseri:
@jzstark the cleanup pr should be more or less finished. If it passes we may be able to just proceed in properly dealing with aeos withot the need for the intermediate state that you suggest

mseri:
@jzstark I think that with the new configure.ml you should have a good example of how to use pkg-config + defaults + test_c to check for libomp and either use the dummy header or the full power. Ping me if you need support with this

jzstark:
@mseri Thanks a lot for your help on improving the Owl configuration! Now that it's stable, I'll try to follow this example to improve the AEOS configuration in similar way.

src/aeos/owl_dune.omp
```diff
@@ -0,0 +1,201 @@
+(copy_files# ext/*)
+(copy_files# core/*)
+
+; copy core/openmp/* with the exception of dummy file
```
mseri: Instead of creating a new file, you can conditionally generate this section and enable aeos when ENABLE_OPENMP (or even better OWL_ENABLE_OPENMP) is set by using the ocaml syntax in dune, e.g. https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/dune#L2

This would be much cleaner than keeping all this copies of stuff lying around and remembering to always keep track of them.

mseri: Depending on the variable you can write in the dune file to either copy the dummy header or the generated ones. Do not set the variable in the dune file as done now (in fact, remove all those lines as they are kind of unneeded now), so that people calling `OWL_ENABLE_OPENMP=1 opam install owl` or `OWL_ENABLE_OPENMP=1 dune build` will have the `aeos` version automatically

src/aeos/config/configure.ml
```diff
@@ -5,51 +5,71 @@
...
 
 
 let get_openmp_libs c =
-  match get_os_type c with
-  | "linux"     -> [ "-lgomp" ]
-  | "linux_elf" -> [ "-lgomp" ]
-  | "macosx"    -> [ "-lomp" ]
-  | "mingw64"   -> [ "-lgomp" ]
-  | _           -> []
+  let enable_openmp = igetenv "ENABLE_OPENMP" in
```
mseri: You just need 
```ocaml
  let enable_openmp = igetenv "ENABLE_OPENMP" in
  if enable_openmp = 0 then []
  else
    match get_os_type c with
    | "linux"     -> [ "-fopenmp" ]
    | "linux_elf" -> [ "-fopenmp" ]
    | "macosx"    -> [ "-Xpreprocessor"; "-fopenmp" ]
    | "mingw64"   -> [ "-fopenmp" ]
    | _           -> []
```

It would fail in line 12 if it is not integer and not 0 or 1.

mseri: Ops, this was meant for the block above, but it's the same here

src/aeos/owl_dune.omp
jzstark: Thanks you so much for your support and help @mseri. After some thoughts on this issue, I think the configuration of AEOS itself is not very well thought of from the beginning. What you have proposed is indeed a much better and systematic solution. However, to reduce the "attack surface" in the Owl configuration issue, I believe it is cleaner for now to just totally disable AEOS from the dune file. Right now I'm bit occupied, and I plan to come back later to give a full review of this module and its configuration.

mseri: As you wish but I think it is actually easier to simply generate the empty file

mseri:
This is what I suggest: https://github.com/owlbarn/owl/compare/master...mseri:aeos-fix-different?expand=1

With this changes, if I install owl normally it does not use aeos, but if I do `ENABLE_OPENMP=1 dune build` or `ENABLE_OPENMP=1 opam install owl # with owl pinned to my branch` it does use aeos (you can test it by disabling the header fix and seeing how the second fails for example)

mseri:
Update: apologies I had pushed the wrong branch, now it should be the correct one

mseri:
Also, by doing this, you make it easy to anybody to enable aeos, devmode or expmode even when using opam

ryanrhymes:
shall we close this one in favour of https://github.com/owlbarn/owl/pull/396 ?

jzstark:
Yes, I think it's better to use #396 rather than this solution. Close it now.

Issue #377, opened 2019-02-15
RFC: split owl-plplot from owl


mseri:
With the appearence of alternative backends like https://github.com/ghennequin/gp and https://github.com/ghennequin/juplot, and with the many limitations of the interfaced provieded for `pplot`, I think it may be worth splitting it out as a separate module.

ryanrhymes:
I agree. I have been thinking about this for some time already. It also helps in making owl's core part more self-contained. I can make a separate repo later. @ghennequin Guillaume, how complete gp and juplot libs are at the moment?

ghennequin:
Both were written in 2 days as I was in a hurry ^^

gp is "complete" in the sense that you can talk to gnuplot directly, so essentially use any command; there are helper functions for commands that I use more often. I had fun with GADTs to try and simplify the syntax, but I'm not entirely sure that I reached the best design. Bear in mind that gnuplot isn't a library, but really a program that you can send commands to via pipes. Not super portable I think, but works very well on linux systems. Historically I have gone for gnuplot because of its high-quality pdf/latex terminals that is great for scientific figures.

Juplot is a super simple module to send figures produced by gnuplot to a jupyter notebook. I doesn't need to grow much bigger than what it currently is, I think.

mseri:
Expect a pr to dunify them soon ;-)

PR #384, opened 2019-02-21
Split out plplot

mseri:
This should help for https://github.com/owlbarn/owl/issues/377


ryanrhymes:
Thank you Marcello, shall we also add the plplot dependency into owl-top. It is better to have 'hub' lib which loads everything for users, in order to give them a relatively complete environment.

ryanrhymes:
btw, I moved the plplot folder into src which also includes all other sub libs.

mseri:
Do you want me to send a separate pr for the owl-top thing or you'll deal with it?

ryanrhymes:
yes, a PR would be nice, thanks 👍

PR #385, opened 2019-02-21
Owl cleanup configure part1

mseri:
* autodiscover openblas via pkg-config if possible (especially useful if you provide a specific PKG_CONFIG_PATH)
* check if lapacke is needed (hopefully enough to fix the build)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,19 @@ script: bash -ex ./.travis-docker.sh
 
 env:
   global:
-    - OCAML_VERSION=4.06
-    - PINS="owl-base:. owl-aeos:. owl-top:. owl-zoo:. owl:."
-    - DISTRO=debian-stable
-  matrix:
     - PACKAGE="owl"
+    - PINS="owl-base:. owl-aeos:. owl-top:. owl-zoo:. owl:."
+
+matrix:
+  include:
+  - os: osx
+    osx_image: xcode10.1
+    env: OCAML_VERSION=4.07
+  - os: linux
+    env: OCAML_VERSION=4.06 DISTRO=debian-stable
+  - os: linux
+    env: OCAML_VERSION=4.07 DISTRO=debian-stable
+  - os: linux
+    env: OCAML_VERSION=4.06 DISTRO=fedora
+  - os: linux
+    env: OCAML_VERSION=4.07 DISTRO=ubuntu
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,10 @@ env:
 
 matrix:
   include:
-  - os: osx
-    osx_image: xcode10.1
-    env: OCAML_VERSION=4.07
+  # we need to use a custom setup instead of docker for osx, postponed
+  # - os: osx
+  #   osx_image: xcode10.1
+  #   env: OCAML_VERSION=4.07
   - os: linux
     env: OCAML_VERSION=4.06 DISTRO=debian-stable
   - os: linux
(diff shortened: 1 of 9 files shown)


mseri:
I don't yet know if this is enoug to fix the build, I am using the PR to check it. The pkg-config setup for openblas works nicely on my mac.

mseri:
Now owl compilation seems to work fine, but it is failing to link eigen. I wonder if some changes in `dune` have to do with that :(

mseri:
I have closed/reopened to restart the last travis build that got stuck initialising the container

mseri:
With the last commit we seem to have ruled out dune issues

mseri:
@ryanrhymes it works!!! Two tests are failing:
```
[ERROR]             lazy evaluation              1   fun01.

[ERROR]             lazy evaluation              7   fun07.

- -- lazy evaluation.001 [fun01.] Failed --
- in /home/opam/.opam/4.07/.opam-switch/build/owl.0.4.0/_build/default/test/_build/_tests/4706AC02-EA57-4497-B19C-887AD80B40D2/lazy evaluation.001.output:
- --------------------------------------------------------------------------------
- ASSERT fun01
- --------------------------------------------------------------------------------
- [failure] Error fun01: expecting
- true, got
- false.
- 
- The full test results are available in `/home/opam/.opam/4.07/.opam-switch/build/owl.0.4.0/_build/default/test/_build/_tests/4706AC02-EA57-4497-B19C-887AD80B40D2`.
- 2 errors! in 3.536s. 1383 tests run.
```

And the runtest is doing something weird:
```
 (cd _build/default/test && ./test_runner.exe)
- tput: No value for $TERM and no -T specified
- stty: 'standard input': Inappropriate ioctl for device
```

But I believe this may be unrelated from this PR. Before merging it we will need a new release of eigen though.

mseri:
eigen release is pending here: https://github.com/ocaml/opam-repository/pull/13506

mseri:
The two failures are actually independent of this PR and are already affecting master: 
https://travis-ci.org/mseri/owl/jobs/497524896

mseri:
@ryanrhymes the new eigen has just been released. I have rebased over the current master and removed the pins to eigen, let's see if it passes now

mseri:
Even with the latest changes from #388 the two lazy tests are failing, see
https://travis-ci.org/owlbarn/owl/jobs/497750331

mseri:
There is surely a large margin of improvement, like using `pkg-config` for all the libraries to avoid having to hack paths into the script and keep changing it manually each time, but I think this could be better done incrementally in separate PRs.

I also think that fixing those failing tests and having a good build on Travis should be a priority, otherwise we may keep inadvertedly adding bug in the repo.

Anyways I beleieve that both things should be dealt with separately from this PR.

mseri:
@ryanrhymes I think the `unit_lazy` tests are failing because we are comparing floats with polymorpic equality. I have added two commits that replace this with an implementation comparing the l1 norm of the difference of the matrices to a tolerance that is just half of the machine epsilon value. This made the test suceed reliably in my Travis test branch.

If you agree that this is the right approach maybe we should consider extending this kind of comparison to the other unit test and replace the equality or add an equivalent comparison function in the ndarray modules.

If you disagree with this approach feel free to remove the last two commit before merging, as I said before I think this is somewhat out of scope for this PR, but I wanted to make sure that this PR could actually solve the compilation issues and give us proper continuous integration.

ryanrhymes:
I agree, it is a good fix. Numerical error has been haunting various owl's unit tests, would be nice if we have a uniformed way.

ryanrhymes:
Hurray to this great PR! A big thank to Marcello 🥇 finally owl's building and installation are moving into a promising direction!

PR #386, opened 2019-02-23
docker: opam install eigen

ehirdoy:
eigen introduced dune.


ehirdoy:
Need some more work.....

mseri:
I think eigen is not compiling on alpine because the compiler was too old or too new but check.
In unbuntu there are still issues wehn lapacke and openblas are used, the solution is to compile a custom openblas and add its path to PKG_CONFIG_PATH

PR #387, opened 2019-02-23
docker: fix eigen installation

ehirdoy:
Other Linux distribution could do the same.


mseri:
With the recent release of eigen 0.1.1 this should no longer be necessary

ehirdoy:
@mseri I tried once https://github.com/owlbarn/owl/pull/386 but it didn't work. Any hint?

PR #391, opened 2019-02-25
added loggamma to Ndarray and Matrix

tachukao:
I've added the function `lgamma` to both Ndarray and Matrix.

Note that like in the `Maths` module, here I'm using `owl_cephes` for the implementation of double-precision `lgamma` and the default c option for single-precision (`lgam` in `owl_cephes`).

I have opted to directly include `owl_cepehs.h` into `owl_ndarray_maths_stub.c`, but I'm not sure if this is the preferred solution. I had to comment out definition of `exp10` in `owl_ndarray_maths_stub.c` because it is already defined in `owl_cephes`. Happy to make changes if you think we should just stick with the native c implementation of `lgamma`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,26 @@ script: bash -ex ./.travis-docker.sh
 
 env:
   global:
-    - OCAML_VERSION=4.06
-    - PINS="owl-base:. owl-aeos:. owl-top:. owl-zoo:. owl:."
-    - DISTRO=debian-stable
-  matrix:
     - PACKAGE="owl"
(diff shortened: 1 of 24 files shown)


ryanrhymes:
This is fine by me if you are confident that two algorithms won't generate significantly different results, otherwise users may get confused by seeing quite different results when switching between precisions.

For the format, it is better to put new lgama function into the group of unary operators rather than appending it to the end of a file, e.g. in .c, .h, and .ml files. This is easier for others to study code structure and maintaining the code.

PR #393, opened 2019-02-26
added lgamma function

tachukao:
I'm reopening the PR for adding lgamma with a cleaner commit histroy.

Here I'm using `owl_cephes`'s `lgam` for the implementation of double-precision and the default c option `lgammaf` for single-precision. 
Note that the scalar `loggamma` function in the `Maths` module also uses `owl_cephes`'s `lgam` function.

Currently the single and double precision `lgamma` gives similar results up to around 8 decimal points.
We might consider including single-precision `cephes` functions in the future.


ryanrhymes:
i re-triggered the travis ci since previous test failed again. the reason is because lyapunov test takes random inputs for testing and the numerical error is still sometimes greater than the threshold we set. i think it is fine, just need to re-trigger ci, setting a very loose threshold might make the test pointless.

the new PR looks very good, thanks 👍

PR #396, opened 2019-03-01
Aeos: infer usage from flags

mseri:
I had some time to polish the approach I was suggesting here https://github.com/owlbarn/owl/pull/383

With this change, to enable aeos/openmp you just have to compile with `OWL_ENABLE_OPENMP=1`. In any other case you will get the usual `owl`. In case the flag is set but the linking fails, the configuration will exit with a readable error message and some information for error reporting.

I think this is simpler to maintain and to use than having multiple copies of build files and having to manually copy them around and maybe commit them by mistake.

We could also add a travis test for owl with OPENMP, EXP and DEV modes enabled in this way, as now it all depends on manually setting these variables in the environment, and no longer on modifying build files.

Many thanks to @jzstark as multiple fixes in this PR come from his work for https://github.com/owlbarn/owl/pull/383


mseri:
Concerning your comments:

1. This is on purpose, instead of having a static dummy file, I let aeos generate the dummy file if the variable is set to 0 (or not set) and do the whole generation + aeos optimization if it is set to 1.

2. This is on purpose, otherwise dune will override whatever I set with 0. The configure scripts now consider a missing variable equivalent to a variable set to 0. You only need to modify them if you want to enable the features.

Issue #394, opened 2019-02-26
Override arch optimizations


rdavison:
Is there any support or any planned support for being able to override the `-march=native flag`?  I am currently trying to support a continuous integration and delivery pipeline which begins by checking out application source code to a virtual machine in the cloud, building executables, saving the executables to a docker image, and then deploying the docker image to another cloud environment.

Unfortunately, since `-march=native` is hard coded in a few different places [1] [2] [3], it means that when my cloud CI provider encounters internal scaling issues due to rapid growth of their platform, they may upgrade their underlying virtual machines with fancier hardware.  In my case, they upgraded to machines with Intel Xeon processors which supported AVX-512 instructions sets.  This led to binaries being compiled with fancy optimizations which later crashed when deployed to our vanilla x86-64 hardware.

For the time being, we are using a fork with those `-march=native` flags set to `-march=x86-64`, but it would be nice if those settings could be overridable somehow.  Any thoughts? Could perhaps the configurator in `dune` be of any use in this situation?

[1]
https://github.com/owlbarn/eigen/blob/master/eigen/dune#L29

[2]
https://github.com/owlbarn/owl/blob/master/src/owl/config/configure.ml#L81

[3]
https://github.com/owlbarn/owl/blob/master/src/aeos/config/configure.ml#L15

mseri:
We can read environment variables, it would not be hard to allow to override thse by reading them out of `OWL_CFLAGS` when set. That would work cleanly both with opam and manually compiling.

@ryanrhymes what is your opinion in this respect?

mseri:
This would probably also provide a fix for https://github.com/owlbarn/owl/issues/354

ryanrhymes:
env variables is a good solution, we can provide default settings, so I agree.

mseri:
Ok, than this is an easy fix. I will send a PR here and to eigen after #396 is merged

mseri:
@rdavison with the latest eigen and the current master branch of owl, you should be able to override all the C and CPP flags by setting the right env variables (refer to owl and eigen's readmes for the names and the default values). Does this solve your issue?

mseri:
@rdavison with he released eigen and owl you can customise it, see 
https://github.com/owlbarn/owl/blob/master/docker/Dockerfile.ubuntu and
https://github.com/owlbarn/owl/blob/master/docker/Dockerfile.ubuntu.arm as examples of use. Let me know if this solves your issues

rdavison:
[...]

PR #397, opened 2019-03-01
config-flags: allow to customise owl and owl_aeos cflags

mseri:
This should help fixing https://github.com/owlbarn/owl/issues/354 and https://github.com/owlbarn/owl/issues/394

Would it make any sense to unify owl and aeos cflags or it is better to keep them separate?
Is there any modification you would like to have for the current default values?

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,5 +20,17 @@ before compilation:
 - `OWL_ENABLE_OPENMP=1`: turn on OpenMP support in core module and
   the automatic parameter tuning (AEOS)
 
+- `OWL_CFLAGS` allows to change the default flags passed to the C targets,
+  it defaults to
+  ```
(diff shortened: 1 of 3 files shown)


ryanrhymes:
It is better to separate owl and aeos. Current default values look good to me, they lead to reasonably good performance on various platform AFAIK.

mseri:
Ok. Thenk I think this PR is ready. Would it make sense to draft a new opam release of owl after this?

ryanrhymes:
Yes, it is about time to make a new release. I might also make sense to jump to 0.5.0 since there are many significant changes in owl. I hope dune release will work this time, I haven't tried since last time it failed due to https://github.com/samoht/dune-release/issues/116#event-2157346068

mseri:
It should work, usually I try to tag and release with `dune-release` and it works ok, but if there are issues (link the nonstandard build rules in eigen) I proceed with `opam-publish` and it works fine.

PR #400, opened 2019-03-11
improve stability of lyapunov gradient tests

tachukao:


mseri:
Thanks! This should make the tests more reliable.

It may be worth for future purposes to annotate (in the commit message or in the code) why you did make this change

PR #402, opened 2019-03-25
add median and sort along in ndarray

dinosaur-lin:


ryanrhymes:
This looks really good to me, thank you Yunfeng! 👍

PR #403, opened 2019-03-28
fix build on archlinux (pkg-config cblas)

vbmithr:


src/owl/config/configure.ml
```diff
@@ -182,6 +182,11 @@ the output of `src/owl/config/configure.exe --verbose`.
 
 let () =
   C.main ~name:"owl" (fun c ->
+      let cblas_conf =
+        let default = { C.Pkg_config.cflags = [] ; libs = ["-cblas"] } in
```
mseri: I think there you meant `-lcblas` but in this case I would leave that empty, as in most cases we want to use the `cblas` that comes from `openblas` instead (and `-lcblas` used to break some of the builds when I was refactoring the build process and trying to fix ubuntu).

mseri: Also, I think you should add a test source with `#include <cblas.h> ...` and use that to detect if the `cblas` flags are actually needed.

ryanrhymes:
@mseri are you happy with the revision, do you think we can merge this now?

mseri:
I think it is fine to start with it, we can add more fine checks on when cblas linking is actually needed in a separate PR if it ever becomes necessary

PR #408, opened 2019-04-14
Fix test warnings

pveber:
I did my best to understand the context of each warning and check there was no bug lurking behind. I only found one case where some code was actually dead (and a test case was consequently ignored). All tests pass after this correction.

### DIFF
--- a/test/unit_algodiff_diff_generic.ml
+++ b/test/unit_algodiff_diff_generic.ml
@@ -27,7 +27,7 @@ module Make
 
 
   (* a module with functions to test *)
-  
+
   module To_test = struct
 
     let dumb () = true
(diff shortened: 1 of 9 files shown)


jzstark:
Thanks for fixing these tests and code format issue. I've tried it with OCaml v4.07.0, and all tests passes.  Looks quite good to me, especially the fix of convolution2D tests I wrote a while ago. Maybe @tachukao could also help to provide some insight on the fix of AlgDiff tests.

ryanrhymes:
Also look very good me. Thank you @pveber for improving the tests. And my apologies for the delay in responding. Great work!

pveber:
Great, thanks both for reviewing this PR!

PR #409, opened 2019-04-18
Add parameter support to Zoo

jzstark:
This PR fixes zoo so that `owl -run` accepts extra parameters now.


ryanrhymes:
This finally fixes the issue of passing in parameters to zoo script. Great work Jianxin! Now Zoo is a complete system ;)

PR #412, opened 2019-06-05
Reorganise Algodiff module

tachukao:
I have substantially reorganised the Algodiff module to make it more readable and easier to add new functions. 

In particular, I have removed the `trace_op`  type and opted instead to use the types `reverse`, `inputs`, and `label` to keep track of the functions in reverse mode.

I have also separated out neural network and linear algebra functions from the `Maths` module into their own separate modules.

Last but not least, I have separated out the definition of new Algodiff operations into a separate file `owl_algodiff_ops.ml`, so that it would be easier to add new operations in the future.

I've tried to split the `owl_algodiff_ops.ml` file further into separate files to improve readability, but it seems a bit difficult, might be something to look into in the future.


jzstark:
This is an awesome PR. The original `owl_algodiff_generic.ml ` file is now restructured into a `Core -> Ops` functor stack. The multiple aspects of an operation such as reverse and backward operation are now organised together. The structure is much cleaner now, especially when adding new operations.

The code in this PR is mostly restructured from existing code. All the unit tests pass, and the newton method, gradient, regression, and mnist examples also work fine. I believe Calvin must have already tested the code on several other AlgDiff examples. 

The only small issue here is about the code format. I think it would be good to follow existing coding style and align the "->" arrows in a match statement if possible. Also, some spacing should be added, e.g. at Line 228 in `src/base/optimise/algodiff/owl_algodiff_generic.ml`.

ghennequin:
@jzstark re formatting: I think it would be awesome to agree on a `.ocamlformat` config file, and have this file in the repo so that we can all use ocamlformat (or `dune build @fmt` before each commit) to produce uniform formatting. Others, any thoughts?

My favourite:
```
profile=janestreet
let-binding-spacing=sparse
```

jzstark:
Agree. The Owl code format loosely follows the [Cornell style](http://www.cs.cornell.edu/courses/cs3110/2011sp/Handouts/style.htm), but a formal format would be indeed quite helpful. I don't have personal preference about formatting as long as it's consistent everywhere.

ryanrhymes:
@tachukao please do, ocamlformat seems working well on owl_ode.

tachukao:
I have just made some extra changes to the algodiff module.
The main changes are a new `Builder` module for building `Algodiff` operations. An example of how to do that can be found in `examples/custom_algodiff_op.ml`.

The main idea behind the Builder module is to classify operations based on the number of inputs and outputs. For example, `Builder.build_siso` allows the user to efficiently build single input single output operations. These builder functions are used to create most of the supported Algodiff operations in `owl_algodiff_ops.ml`.

I've also separated out the `Grad_test` module and the `reverse mode` operations from the main `owl_algodiff_generic.ml` file, hopefully making things more readable.

tachukao:
> @tachukao please do, ocamlformat seems working well on owl_ode.

cool, I'll make a separate pull-request, after this one's done 😄

ryanrhymes:
@tachukao great job! Algodiff now has much greater flexibility 👍

PR #415, opened 2019-06-06
Update examples

pratapsingh1729:
This pull request updates some of the files in the `examples/` folder to work with the current version of Owl.  The changes are as follows:

- Changed `owl_zoo_cmd.ml` such that the `owl` shell command automatically loads the `owl-plplot` package, so that plots can be made using the scripting system.
- Changed `computation_graph_01.ml` and `computation_graph_02.ml` to work with the current functions for writing a computation graph to a dot file.
- Changed `test_log.ml` to remove some type errors.
- Changed `svm.ml` and `regression.ml` to load the `Owl_plplot` module to allow plotting.

PR #416, opened 2019-06-06
Formatted repo using ocamlformat

tachukao:
I've reformatted code in owl using ocamlformat.

Currently, I am ignoring files inside the `examples`, `src/opencl` and `perftest` directory by adding a `.ocamlformat` file in these directories with `disable` written in it. The reason for this is that ocamlformat currently doesn't work well with files with `#!...` written at the top. Hopefully, this will change in the future.

Also, we might consider dunifying the examples.


tachukao:
The latest commit makes sure that we are only disabling ocamlformat for the executable files in `src/opencl/bindings`, but still making sure it works for the rest of the files in `src/opencl`

src/base/compute/owl_computation_type.ml
```diff
@@ -54,37 +60,37 @@ module Make
...
-    | Sequential                    of int array
-    | Uniform                       of int array
-    | Gaussian                      of int array
-    | Bernoulli                     of int array
-    | Init                          of int array * (int -> elt)
-    | Get                           of int array
-    | Set                           of int array
-    | GetSlice                      of int list list
-    | SetSlice                      of int list list
+    | Empty of int array
```
jzstark: This one might be difficult, but do you think it is possible to keep the alignment of each item in `match`?

src/base/core/owl_graph.ml
```diff
@@ -3,101 +3,75 @@
...
+  | PostOrder
+
+type dir =
+  | Ancestor
+  | Descendant
 
 let _global_id = ref 0
-
-
 let id x = x.id
```
jzstark: I think it would be better if these space lines between functions could be kept, but perhaps it's just my personal preference.

src/base/compute/owl_computation_graph.ml
```diff
@@ -5,55 +5,50 @@
...
-    mutable iopair : (attr node * attr node) array;  (* input and output loopback pairs *)
-    mutable iosafe : bool array;                     (* whether it is safe to use unsafe_assign_arr *)
-    mutable random : attr node array;                (* rvs automatically invalidate themselves *)
-    mutable htbl   : (string, attr node) Hashtbl.t;  (* node name to node mapping *)
-    mutable device : device                          (* device-dependent field *)
-  }
-
+  type graph =
+    { mutable name : string
+    ; (* name of the graph *)
```
jzstark: The positioning of `;` looks really weird. I was thinking if it's possible to just keep it at the end of a line.

jzstark:
In general I think using this automatic tool to enforce a format is really helpful, but I do have some minor concerns about the current format, especially considering such a dramatic change. Instead of take all the `janestreet` rules as is, perhaps it would be prudent to fix some aspects of this profile until the code format is in a better shape.  Thanks a lot for this great improvement anyway!

src/base/compute/owl_computation_type.ml
tachukao: so turns out this is a feature that they are considering adding in the future; it seems that it's not so straightforward to do  (see https://github.com/ocaml-ppx/ocamlformat/issues/360)

src/base/compute/owl_computation_graph.ml
tachukao: My latest commit fixes that. I agree that with the comments,  it does look a bit weird; otherwise, I do normally prefer `;` at the front.

src/base/core/owl_graph.ml
tachukao: I don't really see a way of changing that changing .ocamlformat at the moment, but I do like how compact it is at the moment

src/base/compute/owl_computation_type.ml
ryanrhymes: hope this feature can be added soon, it looks better imo

ryanrhymes:
Agree, it is good we can move to a more automated process to enforce the coding style. I think we can merge this PR, then gradually tweak the format to suit our need.

src/base/compute/owl_computation_symbol_sig.ml
```diff
@@ -6,211 +6,217 @@
...
   (** TODO *)
+  val unsafe_assign_arr : arr -> A.arr -> unit
 
-  val assign_arr : arr -> A.arr -> unit
   (** TODO *)
+  val assign_arr : arr -> A.arr -> unit
 
-  val assign_elt : elt -> A.elt -> unit
   (** TODO *)
+  val assign_elt : elt -> A.elt -> unit
```
ryanrhymes: oops, i think the comment went back to the top.

tachukao: oh yea I forgot to mention, I've changed owldoc to read the doc with the comments at the above the function and type definitions...
the problem here is even if I set the doc-comments to come after in ocamlformt, it only comes after the functions definitions but not the type definitions (see https://github.com/ocaml-ppx/ocamlformat/issues/556) 

If we really want the comments to come after the function definitions, I can go back to owldoc code, but that will probably take some more time.

ryanrhymes: can ocaml's own doc builder recognise this if the comment is above the function?

mseri:
Please wait to merge this until #415 has been merged, as that would likely need to be re-done manually from scratch, while this is generated automatically

src/base/compute/owl_computation_symbol_sig.ml
tachukao: So that's how I've modified owl doc's `apidoc.ml` this afternoon (see https://github.com/owlbarn/owl_guide/blob/master/bin/apidoc.ml)

tachukao:
> Please wait to merge this until #415 has been merged, as that would likely need to be re-done manually from scratch, while this is generated automatically

#415 seems to only relate to files in `examples` which is currently being ignored by ocamlformat, as the formatter is currently having difficulty dealing with `#!/usr/bin/env ow` at the top and `#zoo...`

mseri:
Oh ok, I had missed that. It's a pity that we cannot apply the style uniformly though

ryanrhymes:
@tachukao @jzstark @mseri I feel like this PR is too huge w.r.t the number of files it has changed, this will impact serval people's work at the same time immediately. However I do think we need this mechanism to enforce coding style. So I suggest we adopt the changes incrementally and gradually.

- Can we first have a single PR which only introduces ocamlformat into Owl with an empty profile (i.e. owl's own profile) similar to janestreet one, so no source files will be actually changed in this PR.

- After ocamlformat is incorporated in Owl, we can gradually add more rules into owl's profile little by little, in separate PRs so we can review each rule and reach the agreement. Currently, it changes so much so it is really difficult for me to review everything all at once and estimate the impacts.

- ocamlformat itself is also evolving, in my proposed way, we can wait for some features to be implemented in ocamlformat before we actually introduce some style rules.

Do you think this makes sense?

jzstark:
Agree. A gradual approach with one rule added at a time is a much safer approach IMO. Should there be any unknown error after a rule is introduced, reverting this change would be less harmful.

tachukao:
> @tachukao @jzstark @mseri I feel like this PR is too huge w.r.t the number of files it has changed, this will impact serval people's work at the same time immediately. However I do think we need this mechanism to enforce coding style. So I suggest we adopt the changes incrementally and gradually.

@ryanrhymes @jzstark @mseri 

I think doing it gradually makes a lot of sense. 

The only problem is that ocamlformat doesn't really work by individual features. As far as I'm aware, you apply all features when you call ocamlformat. The only thing you can change are the different options for each feature. So if you format the code with an empty `.ocamlformat` file, you will be applying all default options.

So perhaps the way to do things gradually is to disable it for most subfolders (We can do this by adding an `.ocamlformat` file in these subfolders); and only apply ocamlformat to specific folders or projects  when we feel that they are ready. 
The problem with this approach however will be creating inconsistencies in formatting across files, which may create problems for owldoc.

For now, I've reverted my pull-request to before applying ocamlformat to the entire code base and only added the relevant `.ocamlformat` files.

ryanrhymes:
OK, this sounds good to me. We can try to request a feature in ocamlformat so that we can gradually include rules.

Issue #116, opened 2017-10-13
Owl crash when environment variable HOME is not defined


rixed:
Owl should probably not crash depending on the environment.
But beyond that, I don't think it is OK to create any subdirs in $HOME by default.
I use owl to do some matrix computation in a program that runs as root in some restricted system
where creating directories here and there is frowned upon.

As I understand it from a quick glance, those directories are needed by the Zoo subsystem, which is an improved toplevel. Maybe this whole Zoo subsystem should not be included in the core lib owl.cm{x}a at all, but only in owl_zoo.cm{x}a?

ryanrhymes:
@rixed Thanks for reporting it, you comments make perfectly sense. Zoo subsystem has already been separated out from core library, but the code trying to make sub directory still remains in the `owl_dataset` module. I will move it into zoo or find a better solution for storing local data asap.

PR #421, opened 2019-06-18
Do not create .owl folder when loading owl library

mseri:
I believe that libraries should not come with unexpected surprises, in particular I think that the main owl library should not create files in the system and instead file creation should be triggered by the user action (like downloading something from the zoo or a dataset). The creation of .owl/dataset and .owl/zoo has been moved to the zoo cli and the Dataset.load_data_path functions instead.

This change will also allow to run tests (at least some of them) inside
the opam sandboxed envirionment. In fact this is how I found out about
this.

Note that checking the existence of a folder and then creating it is racy (imagine you are downloading something in the zoo in a terminal and getting two datasets from two different instances of utop or jupyter).

The parenthesis around the try are because ocaml was associating all the rest of the code to the with block.

### DIFF
--- a/bin/owl_bin_zoo.ml
+++ b/bin/owl_bin_zoo.ml
@@ -17,6 +17,15 @@ let main args =
   (* initialise logger *)
   Owl_log.set_color true;
   Owl_log.(set_level DEBUG);
+
+  (* initialise owl zoo working environment *)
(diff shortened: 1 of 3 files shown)


rixed:
Related to #116.

PR #427, opened 2019-07-09
Fix exponential rvs

### LINKED ISSUE #426: super odd bug in `exponential_rvs`
Super weird:
```
let exp_rv = Owl_stats.exponential_rvs
Mat.init 1000000 1 (fun _ -> exp_rv 1.0) |> Mat.mean'
```
consistently returns ~1.03 (very, very significantly above 1.0).
In contrast, hand-coding the draw from the exp. distribution, using the SAME CODE as in [the owl source code](https://github.com/owlbarn/owl/blob/master/src/base/stats/owl_base_stats_dist_exponential.ml), i.e:
```
let exp_rv lambda =
  let u = Random.float 1.0 in
  let s = -1. /. lambda in
  s *. log1p (-.u)

Mat.init 1000000 1 (fun _ -> exp_rv 1.0) |> Mat.mean'
```
returns 1.000 as it should. WTH :) Any idea?

ghennequin:
Fixes #426  and fixes the inconsistency in the meaning of `lambda` for the exponential distribution between owl base and owl c; since it usually refers to the inverse mean, I modified zaggurat, not base.

### DIFF
--- a/src/owl/stats/owl_stats_ziggurat.c
+++ b/src/owl/stats/owl_stats_ziggurat.c
@@ -54,7 +54,7 @@ inline double exponential_rvs (double lambda) {
 // init the internal state for exponential prng
 void std_exponential_rvs_init ( ) {
   double de = 7.697117470131487;
-  const double m2 = 2147483648.0;
+  const double m2 = 4294967296.0;
   double te = 7.697117470131487;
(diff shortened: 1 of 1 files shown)


ryanrhymes:
Guillaume thank you for fixing this. Sorry for the delay, I have been travelling recently.

PR #428, opened 2019-07-11
Count-Min and Heavy-Hitters sketches

pratapsingh1729:
This PR implements the Count-Min sketch, and a data structure for finding heavy-hitters built on top of it.  The sketches are described in full detail in [Cormode and Muthukrishnan](http://dimacs.rutgers.edu/~graham/pubs/papers/cm-full.pdf).  The Count-Min sketch provides an approximate frequency table whose space and time complexity are constant in the size of the stream - they depend only on the desired accuracy of the sketch.  The Heavy-Hitters sketch provides a simple approximate data structure for finding the _k_-heavy-hitters among a stream or collection of elements (the _k_-heavy-hitters are those elements whose relative frequency in the input is at least 1/k), with space and time complexity dependent only on the desired accuracy of the sketch and the number of heavy-hitters.  For more details of how the sketches work or the guarantees they provide, see the comments in the `.mli` files or in the Cormode and Muthukrishnan paper.  The Count-Min sketch has been used in networking ([1](https://dl.acm.org/citation.cfm?id=1863207), [2](https://ieeexplore.ieee.org/abstract/document/6006023), and others), and existing implementations include Twitter's [algebird library](https://github.com/twitter/algebird).

This implementation works in streaming scenarios where data is arriving continuously and estimates are required at any time.  Additionally, the implementation is functorized over the choice of table to use for the Count-Min sketch.  Two possible tables are provided, one using OCaml native arrays and one using the Owl ndarray.  The native array is generally about 25% faster and uses somewhat less memory, so should generally be preferred.

I also made an example `countmin_texts.ml` to demonstrate the use of the data structure.  It finds the 100-heavy-hitters (the words appearing at least 1% of the time) in the `news.txt` corpus from https://github.com/ryanrhymes/owl_dataset.  The corpus contains approximately 55 million words of English text from online news articles.

To benchmark the data structure, I compared its performance on the `news.txt` corpus to that of a naive `Hashtbl`-based frequency table.  On a 64-bit 2017 MacBook Pro, the `Hashtbl` implementation required approximately 3.1 million heap words.  An example Count-Min sketch with epsilon = 0.0001 and delta = 0.001, achieving a mean relative error of less than 0.2% in estimating the frequencies of the 1000-heavy-hitters, required about 100,000 heap words.  To illustrate the performance of the sketch, I generated a [plot](https://github.com/pratapsingh1729/owl/blob/master/owl-sublinear/plots/tradeoff_by_phis.pdf) showing the tradeoff between memory usage and accuracy of the sketch.  The data structure provides good accuracy with low resource usage and thus has many potentially interesting applications.


ryanrhymes:
@pratapsingh1729 awesome PR 💯 count-min is a very valuable addition to owl base lib.

PR #431, opened 2019-07-28
Fix crash when printing error message

rgrinberg:
string_of_sexp is the wrong function to convert sexps of strings. This
function attempts to parse an Atom sexp into a single string. Instead,
Sexp.to_string will render the sexp.


ryanrhymes:
Ah, thanks for fixing this!

Issue #347, opened 2019-01-17
Numerical robustness of balance_last


pveber:
The following code fails for me:
```ocaml
let u = [|7.29051884672767705e-28; 7.40891406114041623e-19; 7.8974585653914162e-27;
    6.74931722499952811e-27; 1.96567242595806183e-18; 1.99759418829810833e-09;
    9.00957709439775126e-18; 7.69975469891059e-18; 5.44630568871032146e-27;
    5.53475162082589e-18; 7.14562067029471901e-27; 7.26166281614767e-18;
    3.40581066900184642e-26; 1.76422660684586938e-18; 1.79287696675020245e-09;
    8.08625861896575203e-18; 6.91066930614189608e-18; 9.84019855846523357e-10;
    0.999999958920365861; 4.51021371572784489e-09; 3.85451380633076905e-09;
    5.57697548253416743e-18; 5.6675434090004686e-09; 2.84750345476462241e-17;
    2.43353022074415654e-17; 6.00096499322304467e-18; 6.09841835106380515e-09;
    2.7505171256294709e-17; 2.35064387154075702e-17; 6.62930524914401317e-27;
    6.73696260890209106e-18; 3.03851424471216853e-26; 2.60227107878253518e-26;
    8.19698787626752916e-18; 8.33010379906068e-09; 3.75705500032367818e-17;
    3.21085014652355746e-17; 4.22071117495635736e-26; 4.28925389538807169e-17;
    1.04857628916290599e-25; 8.96133096150629102e-26; 3.89639425302617225e-26;
    3.95967019196240566e-17; 1.31783685040006345e-25; 1.12624825529808348e-25;
    1.2664323265349806e-28; 1.28699869924975191e-19; 5.80463937719803923e-28;
    4.96075441996169876e-28; 7.71899620480556344e-18; 7.84434972714204e-09;
    3.5379695143730131e-17; 3.02361555328812682e-17; 7.34588525560146473e-27;
    7.46517960274637435e-18; 1.49279215819264218e-25; 1.27576836720746576e-25;
    2.33797209106778106e-26; 2.37593985643049289e-17; 1.07159969713535906e-25;
    9.15809335252883654e-26|];;

Owl.Utils.Array.balance_last 1. u;;
```
with an assertion error on this code:
```ocaml
let balance_last mass x =
  let k = Array.length x - 1 in
  let q = ref mass in
  Array.mapi (fun i a ->
    assert (!q >= 0.);
    if i < k then (
[...]

PR #434, opened 2019-08-16
Stats_dist: don't use balance_last for multinomial distribution (#347)

pveber:
As described in #347, balance_last raises exceptions in case where
numerical inaccuracy prevent summation to 1 of input
probabilities. The comments there argue that not only the
transformation performed by balance_last, but is also not really
necessary. See notably this comment:

As I see it, owl's current implementation is very close to the one in
the GSL. It computes the sum of input probabilities (norm variable),
and it seems to me that it doesn't matter that these sum to 1. Playing
with gsl confirms it:

```ocaml
# Gsl.Randist.multinomial rng ~n:100000 ~p:[|1. ; 2.|];;
- : int array = [|33251; 66749|]
```

This seems very natural to me, and much less confusing and error prone
than ignoring the last element of the probability vector.

So in summary, it would be enough to remove the call to balance_last
in multinomial_rvs:

```ocaml
let multinomial_rvs n ~p =
  let k = Array.length p in
  let _p = Genarray.create float64 c_layout [|k|] in
  let _s = Genarray.create int32 c_layout [|k|] in
  Owl_utils.Array.balance_last 1. p |>
  Array.iteri (fun i a -> Genarray.set _p [|i|] a);
  _multinomial_rvs ~k ~n ~p:_p _s;
  Array.init k (fun i -> Genarray.get _s [|i|] |> Int32.to_int)
```
The same is true IMO for the pdf:
```ocaml
# Gsl.Randist.multinomial_pdf ~n:[|3;6|] ~p:[|1. ; 2.|];;
- : float = 0.273129096174363206
# Gsl.Randist.multinomial_pdf ~n:[|3;6|] ~p:[|0.333333333333333 ; 0.6666666666666|];;
- : float = 0.273129096174363706
```

### DIFF
--- a/src/owl/stats/owl_stats.mli
+++ b/src/owl/stats/owl_stats.mli
@@ -492,11 +492,10 @@ from ``n`` trials. The probabilty mass function is as follows.
 .. math::
   P(x) = \frac{n!}{{x_1}! \cdot\cdot\cdot {x_k}!} p_{1}^{x_1} \cdot\cdot\cdot p_{k}^{x_k}
 
-``p`` is the probabilty mass of ``k`` categories. If the elements in ``p`` do
-not sum to 1, the last element of the ``p`` array is not used and is replaced
-with the remaining probability left over from the earlier elements.
+``p`` is the probabilty mass of ``k`` categories. The elements in ``p`` should
+all be positive (result is undefined in that case), but they don't need to sum
+to 1. For implementation details, refer to :cite:`davis1993computer`.
 
-For implemantation, refer to :cite:`davis1993computer`.
  *)
 
 val multinomial_pdf : int array -> p:float array -> float
--- a/src/owl/stats/owl_stats.mli
+++ b/src/owl/stats/owl_stats.mli
@@ -493,9 +493,9 @@ from ``n`` trials. The probabilty mass function is as follows.
   P(x) = \frac{n!}{{x_1}! \cdot\cdot\cdot {x_k}!} p_{1}^{x_1} \cdot\cdot\cdot p_{k}^{x_k}
 
 ``p`` is the probabilty mass of ``k`` categories. The elements in ``p`` should
-all be positive (result is undefined in that case), but they don't need to sum
-to 1. For implementation details, refer to :cite:`davis1993computer`.
-
+all be positive (result is undefined if there are negative values), but they 
+don't need to sum to 1: the result is the same whether or not ``p`` is normalized. 
+For implementation details, refer to :cite:`davis1993computer`.
  *)
 
 val multinomial_pdf : int array -> p:float array -> float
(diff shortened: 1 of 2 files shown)


src/owl/stats/owl_stats.mli
```diff
@@ -492,11 +492,10 @@ from ``n`` trials. The probabilty mass function is as follows.
 .. math::
   P(x) = \frac{n!}{{x_1}! \cdot\cdot\cdot {x_k}!} p_{1}^{x_1} \cdot\cdot\cdot p_{k}^{x_k}
 
-``p`` is the probabilty mass of ``k`` categories. If the elements in ``p`` do
-not sum to 1, the last element of the ``p`` array is not used and is replaced
-with the remaining probability left over from the earlier elements.
+``p`` is the probabilty mass of ``k`` categories. The elements in ``p`` should
+all be positive (result is undefined in that case), but they don't need to sum
```
mseri: What do you mean with `result is undefined in that case`?

mseri:
Thanks for the PR!

mseri:
Would it be worth to log a warning (containing the value of the sum) if the probabilities do not sum to 1? In this way we can still inform the user that something may be fishy

src/owl/stats/owl_stats.mli
pveber: Woops, sorry the formulation is very unclear. I meant: "result is undefined if not".

Well there's a tradeof here: either we check that all elements in `p` are positive and raise `Invalid_argument` if not; either we check nothing but don't give any guarantee on the result if the precondition is not verified (I have no idea what the function does in that case, although I'm confident it terminates). The thing is that checking `p` at each call can be very costly, and for large scale simulation code you wouldn't want to do that. Another possibility would be to introduce a private type of float array with only positive values, so that the check can be done once and for all.

pveber:
I'm not too fond of logs, but anyway I think there is nothing fishy in providing an unnormalized set of (positive) weights: the underlying algorithm is just as good in that case, and it can be very convenient not to be forced to normalize the weights (and that's what the GSL does). In case you'd agree with that design, I should probably explain better in the doc :).

pveber:
PS: I updated the initial message, it seems I had failed the intended copy/paste

ryanrhymes:
Thanks for fixing this. The design looks very good to me. Logging can be supressed by increasing the logging level in owl so it should not be a problem. @mseri can we merge this?

src/owl/stats/owl_stats.mli
pveber: I added a commit to improve the description in the doc.

mseri:
I like the latest docs... I was just commenting about the wording when you committed :)

pveber:
Alright, thanks @mseri @ryanrhymes !

PR #438, opened 2019-09-02
Parallel countmin sketch

pratapsingh1729:
This PR adds functionality to the count-min sketch to allow for duplication and merging of sketches.  Two count-min sketches with the same parameters can be merged, giving a sketch whose frequencies are the sum of the frequencies in each individual sketch.  This can be used to process multiple parts of a dataset in parallel, by initializing multiple sketches for each part and then merging at the end.

This PR adds functions `init_from` (initialize a new sketch with the same parameters as the given sketch) and `merge` (merge two sketches with the same parameters) to the API for the `Countmin_sketch` module.  The functionality is illustrated by the example `countmin_distributed.ml`, which processes a data set using both a single sketch and two separate sketches, and shows that they produce the same result.


kayceesrk:
This is great. I can now speedup count-min sketches using multicore :-D

PR #439, opened 2019-09-06
Lazify Algodiff Functions

tachukao:
1) Currently, Algodiff operations are rebuilt each time that it is called. In this pull request, I have lazified these operations to improve efficiency.
2) I have also refactored the code and added .mli and signature files


ryanrhymes:
Very neat! The refactored code also looks much better. Is it possible to add a couple of unit tests for the lazy evaluation (can be a separate PR)?

ryanrhymes:
Another thing is, if you want owl to build doc out of the code, the inline doc should be below the function signature ... but this is easy to fix in another PR.

PR #440, opened 2019-09-10
forward mode algodiff test + dimension test for jacobian-vector products

tachukao:
several updates

1) added forward-mode gradient test (tested using reverse-mode gradients)
2) fixed several forward-mode gradient implementations (e.g. l2norm', l2norm_sqr',...)
3) can now differentiate through`jacobianTv`
4) can now differentiate through`concatenate`
5) check that for jacobian-vector products, vector has the right dimensions (`v` should have the same dimension as that of the output in reverse-mode and as that of the input in forward-mode)

TODO:
1) there are some algodiff functions that cannot be differentiated repeatedly (e.g., of_arrays)
2) forward-mode gradients for qr, lq, and split still need to be implemented


mseri:
Concerning 1., why is that?

tachukao:
@mseri currently functions like of_arrays and of_rows do not compare tags when establishing whether to do the reverse, forward, or normal mode computation

For example, these functions would not know how to properly deal with the case that an array element like DR ( DR (...),  ...)

Need to do some book keepings like the ones done in concatenate in this pull request

src/base/optimise/algodiff/owl_algodiff_generic.ml
```diff
@@ -82,41 +84,63 @@ module Make (A : Owl_types_ndarray_algodiff.Sig) = struct
...
+        let s = A.shape x in
+        let d = Array.length s in
+        if d > 2
+        then `arr
+        else if s.(0) = 1
+        then `row s.(1)
+        else if s.(1) = 1
+        then `col s.(0)
+        else `mat
+      | _ -> assert false
```
mseri: Can you raise a meaningful error here or this surely cannot happen?

tachukao: Yea I think with the definition of primal’, this surely cannot happen because it’s recursively extracting the primal... but yea we probably should design better helper functions to distinguish between these types

PR #441, opened 2019-09-15
expose expand and squeeze in ndarray type

tachukao:
This will be useful for writing functors that take input module type `Owl_types_ndarray_algodiff.Sig`.


ryanrhymes:
Tests seem broken.

tachukao:
oh yea I forgot to add these functions to `src/base/dense`, the latest commit should fix it.

PR #442, opened 2019-09-17
ocamlformat algodiff + separate it out

tachukao:
summary
1. set up ocamlformat files such that only files in the `algodiff` folders are formatted
2. make sure the doc strings come after not before using ocamlformat
3. separated out the algodiff folder from under `src/base/optimise/`, it is now `src/base/algodiff`


tachukao:
latest commit align pattern matching cases horizontally

PR #449, opened 2019-10-10
Expose Eigen's Direct and Iterative Sparse System Solvers

steelet41:
See the [eigen] pull request for a comment about efficiency.

I added a dependency on Core for access to the Option module. If you would prefer to remove that and use pattern matching or a home-spun Option.map, that is also fine.

I also exposed eigen's rows() and cols() for convenience and as a defensive measure when setting the .m and .n properties of the solution matrices.

I did find one bug in the transpose() function, the dimensions were not being set correctly.

### DIFF
--- a/src/owl/dune
+++ b/src/owl/dune
@@ -185,7 +185,7 @@
  (flags
   :standard
   (:include ocaml_flags.sexp))
- (libraries core ctypes ctypes.stubs eigen owl-base stdlib-shims))
+ (libraries ctypes ctypes.stubs eigen owl-base stdlib-shims))
 
 (rule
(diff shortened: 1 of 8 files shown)


mseri:
I agree with @tachukao, it is great to expose these functions but we should minimize the dependencies. If we want to avoid rewriting the mapping function, I’d rather add stdcompat as backward compatibility dependency and use Option.map from the new Stdlib exposed through it

mseri:
For what concerns error reporting, that would be also very welcome but i think could be postponed! We can expose it in eigen in a separate PR and then add it here at that stage

ryanrhymes:
I close this since there hasn't been any progress ... reopen if necessary.

PR #451, opened 2019-11-05
function to build aiso operations + algodiff gradients for CARE

tachukao:
1. Added a function `build_aiso` that can be used to build  algodiff operation that takes an array input and produces a single output. This is useful for building operations that take many arguments.

2. One such example is the  continous-time algebraic Ricatti equation, which takes four inputs and produces a single output. I've added `Owl.Algodiff.D.Linalg.care` using  `build_aiso`. 

FUTURE: 
1. use `build_aiso` to build operations such as `concatenate` 
2. `Owl.Algodiff.D.Linalg.dare`

### DIFF
--- a/src/base/algodiff/owl_algodiff_ops_builder.ml
+++ b/src/base/algodiff/owl_algodiff_ops_builder.ml
@@ -362,41 +362,40 @@ module Make (Core : Owl_algodiff_core_sig.Sig) = struct
         (fun (i, t, m, idxs) x ->
           match m, x with
           | _, F _ | _, Arr _ -> succ i, t, m, idxs
-          | `n, DR (_, _, _, _, t', _) -> succ i, t', `r, [ i ]
-          | `f, DR (_, _, _, _, t', _) ->
+          | `normal, DR (_, _, _, _, t', _) -> succ i, t', `reverse, [ i ]
+          | `forward, DR (_, _, _, _, t', _) ->
             if t' > t
-            then succ i, t', `r, []
+            then succ i, t', `reverse, []
             else if t' = t
-            then failwith "clash"
-            else succ i, t, `f, idxs
-          | `r, DR (_, _, _, _, t', _) ->
+            then failwith "error: forward and reverse clash on the same level"
+            else succ i, t, `forward, idxs
+          | `reverse, DR (_, _, _, _, t', _) ->
             if t' > t
-            then succ i, t', `r, []
+            then succ i, t', `reverse, []
             else if t' = t
-            then succ i, t', `r, i :: idxs
+            then succ i, t', `reverse, i :: idxs
             else succ i, t, m, idxs
-          | `n, DF (_, _, t') -> succ i, t', `f, [ i ]
-          | `f, DF (_, _, t') ->
+          | `normal, DF (_, _, t') -> succ i, t', `forward, [ i ]
+          | `forward, DF (_, _, t') ->
             if t' > t
-            then succ i, t', `f, []
+            then succ i, t', `forward, []
             else if t' = t
-            then succ i, t', `f, i :: idxs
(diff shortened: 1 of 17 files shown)


src/base/compute/owl_computation_operator.ml
```diff
@@ -502,6 +502,10 @@ module Make
     trans |> ignore; typ |> ignore;
     raise (Owl_exception.NOT_IMPLEMENTED "owl_computation_operator.linsolve")
 
+  let care ?(diag_r = false) _a _b _q _r =
+    diag_r |> ignore;
+    raise (Owl_exception.NOT_IMPLEMENTED "owl_computation_operator.linsolve")
```
mseri: linsolve?

src/base/algodiff/owl_algodiff_ops_builder.ml
```diff
@@ -347,4 +347,85 @@ module Make (Core : Owl_algodiff_core_sig.Sig) = struct
...
+            else if t' = t
+            then succ i, t', `f, i :: idxs
+            else succ i, t, `f, idxs
+          | `r, DF (_, _, t') ->
+            if t' > t
+            then succ i, t', `f, []
+            else if t' = t
+            then failwith "clash"
+            else succ i, t, `r, idxs
+          | _ -> assert false)
```
mseri: Is this really impossible?

src/base/algodiff/owl_algodiff_ops_builder.ml
```diff
@@ -347,4 +347,85 @@ module Make (Core : Owl_algodiff_core_sig.Sig) = struct
...
+      Array.fold_left
+        (fun (i, t, m, idxs) x ->
+          match m, x with
+          | _, F _ | _, Arr _ -> succ i, t, m, idxs
+          | `n, DR (_, _, _, _, t', _) -> succ i, t', `r, [ i ]
+          | `f, DR (_, _, _, _, t', _) ->
+            if t' > t
+            then succ i, t', `r, []
+            else if t' = t
+            then failwith "clash"
```
mseri: Here and below should we be more informative?

src/base/compute/owl_computation_operator.ml
tachukao: thanks for pointing this out

src/base/algodiff/owl_algodiff_ops_builder_sig.ml
```diff
@@ -109,4 +109,13 @@ module type Sig = sig
   end
 
   val build_piso : (module Piso) -> t -> t -> t
+
```
mseri: This whole module should really be documented...

mseri: NB: I am not suggesting to do it as part of this PR

src/base/algodiff/owl_algodiff_ops_builder.ml
tachukao: you're quite right this wouldn't happen

src/base/algodiff/owl_algodiff_ops_builder.ml
tachukao: I've changed it to `error: forward and backward clash at the same level` as above

src/base/algodiff/owl_algodiff_ops_builder_sig.ml
ryanrhymes: While becoming more flexible and powerful, Algodiff is definitely getting more complicated. It deserve a big chapter in the tutorial book, details its design and application :)

tachukao: yea definitely!

PR #452, opened 2019-11-05
Add gauss solver to linalg module in owl-base

jzstark:
This is an initial implementation of Gauss-Jordan elimination algorithm in the linalg module of owl-base. This is still an on-going work and there are several  aspects that need to be further improved:
- follow the existing specification of Owl such as function naming, or perhaps this function should be placed in the [dense module of owl-base](https://github.com/owlbarn/owl/blob/master/src/base/dense/owl_base_dense_ndarray_generic.ml#L5006);
- support for different number types, including complex number;
- more unit test cases;
- code format checking.


src/base/linalg/owl_base_linalg_generic.ml
```diff
@@ -79,4 +81,117 @@ let is_hermitian x =
...
+              big := v;
+              irow := j; icol := k;
+            )
+          )
+        done
+      )
+    done;
+    ipiv.(!icol) <- ipiv.(!icol) + 1;
+
+    (* TODO: how to swap elegantly? *)
```
mseri: I don't think there is a better way

src/base/linalg/owl_base_linalg_generic.ml
```diff
@@ -7,6 +7,8 @@ type ('a, 'b) t = ('a, 'b) Owl_base_dense_ndarray_generic.t
 
 module M = Owl_base_dense_ndarray_generic
 
+open Owl_exception
```
mseri: This is used very rarely, I think it is better to name it when needed instead of a toplevel open

test/unit_linalg_solver.ml
```diff
@@ -0,0 +1,92 @@
+open Owl
+open Owl_exception
+
+module N = Owl_base_dense_ndarray.D
+module L = Owl_base_linalg_generic
+
+(* define the test error *)
+
+let approx_equal a b =
```
mseri: Maybe we should put this in `Maths` instead of redefining it over and over in different modules

jzstark: Thanks! I replace the implementation with `approx_equal` function in the dense module.

src/base/linalg/owl_base_linalg_generic.ml
jzstark: Thanks for reviewing! I've updated the code accordingly.

Issue #337, opened 2018-12-12
Slicing operations fail on large matrices


ejaasaari:
Using owl from the current master branch, slicing matrices that have more than 2^31 elements seems to cause reads from invalid memory addresses. For example, this snippet mmaps an nxn matrix and writes ones to the last row when n = 46340 (< sqrt(2^31)):

```ocaml
module MX = Owl.Dense.Matrix.S

let mmap_mat sz : MX.mat =
  let fh = Unix.openfile "test" [ Unix.O_RDWR; Unix.O_CREAT ] 0o644 in
  Unix.map_file fh Float32 C_layout true sz

let () =
  let n = 46340 in
  let x = mmap_mat [|n; n|] in
  MX.set_slice [[n - 1; n - 1]] x (MX.ones 1 n);
  MX.get_slice [[n - 1; n - 1]] x |> MX.print ~max_col:10
```

For n = 46340 this works as expected:
```
$ ./a.out

   C0 C1 C2 C3 C4     C46335 C46336 C46337 C46338 C46339
R0  1  1  1  1  1 ...      1      1      1      1      1
```

However, changing to n = 46341 (> sqrt(2^31)) causes a segfault:
```
$ ./a.out
[1]    6160 segmentation fault  ./a.out
```

mseri:
Is this due to the fact that Owl uses the non tail-recursive `List.map` here: https://github.com/owlbarn/owl/blob/master/src/owl/core/owl_slicing.ml#L112 ?

ryanrhymes:
Fixed in https://github.com/owlbarn/owl/pull/458

Issue #457, opened 2019-11-20
SIGSEGV


zbroyar:
Hi,

I have a pretty big matrix (763029x22716) of float64. It is stored in a file and mapped into memory with Unix.map_file (please, see [the gist](https://gist.github.com/zbroyar/3a4f768ce09498282a4f308d8ad5fa6c) for implementation details)

When I try to get a slice of the matrix the program dies with SIGSEGV. The actual code looks like this:
```
  let data = genarray_of_array2 @@ CMMap.connect ~cols:22716 fn in
  let r,c = Mat.shape data in
  printf "Mat.shape = [%d; %d]\n%!" r c;
  let start = 0 and stop = 130000 and i = 1 in
  let slice = Mat.get_slice [[start; stop]; [i]] data (* dies here *)
```

The `connect()` function is pretty simple:
```
  let connect ?rows ?cols fname =
    let rows,cols = determine_dims kind fname (rows,cols) in
    let mfile = openfile fname [O_RDWR] 0o600 in
    let data =
      array2_of_genarray @@
      Unix.map_file mfile kind layout true [|rows; cols|] in
    Hashtbl.add tbl data mfile; data
```

Backtrace points to [this line](https://github.com/owlbarn/owl/blob/master/src/owl/core/owl_slicing_basic_impl.h#L55) of Owl code:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  c_float64_ndarray_get_slice_2 (p=<optimized out>) at src/owl/core/owl_slicing_basic_impl.h:55
55      src/owl/core/owl_slicing_basic_impl.h: No such file or directory.
[Current thread is 1 (Thread 0x15367d9ba740 (LWP 30887))]
(gdb) bt
#0  c_float64_ndarray_get_slice_2 (p=<optimized out>) at src/owl/core/owl_slicing_basic_impl.h:55
#1  0x0000561562b9a905 in c_float64_ndarray_get_slice (p=p@entry=0x561563cd5d20) at src/owl/core/owl_slicing_basic_impl.h:196
#2  0x0000561562b9a9f4 in stub_float64_ndarray_get_slice (vX=23322601421440, vY=23322601419880, vZ=<optimized out>) at src/owl/core/owl_slicing_basic_impl.h:252
#3  0x000056156223cac9 in camlOwl_slicing_basic__fun_3583 () at src/owl/core/owl_slicing_basic.ml:24
[...]

PR #458, opened 2019-11-21
Fix issue #337 and #457

ryanrhymes:
This PR changes number type from int32 to int64 in underlying indexing and slicing function. This change affects many c code in owl core, therefore needs careful examination before merge.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,3 +64,7 @@ release:
 	dune-release opam pkg -p owl-zoo
 	dune-release opam pkg -p owl-top
 	dune-release opam submit -p $(PKGS)
+
+.PHONY: cloc
+loc:
+	cloc src/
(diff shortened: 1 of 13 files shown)


mseri:
Maybe we should use intnat as well. It is defined by the ocaml runtime and chooses between int and long int depending on the arch support: https://github.com/ocaml/ocaml/blob/fe31a0179dd0575f217dcde11ca4dabcc2673c57/runtime/caml/config.h#L127

It is used pervasively in the mmap code in ocaml and would prevent breakage in 32bit arch. 

What do you think?

jzstark:
But perhaps sticking with one explicit type would potentially reduce bug? I'm thinking how widely is Owl used on 32bit machines.

PR #465, opened 2019-12-03
Owl-base linalg routines: LU decomposition 

jzstark:
This PR initially adds LU decomposition in owl-base, including lu-based linear solver, `det`, and `inv` routines. It also contains solver for tridiagonal matrix. The implemented routines are tested with simple unittest cases.

### DIFF
--- a/src/base/linalg/owl_base_linalg_generic.ml
+++ b/src/base/linalg/owl_base_linalg_generic.ml
@@ -277,24 +277,43 @@ let lu a =
   done;
   l, lu, indx
 
-(*
 
 let lu_solve_vec a b =
   (*TODO: check shape; b and x are vectors *)
+  assert (Array.length (M.shape b) = 1);
   let n = (M.shape a).(0) in
-  if ((M.shape b).(0) != n || (M.shape x).(0) != n) then
+  if ((M.shape b).(0) != n) then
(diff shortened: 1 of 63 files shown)


src/base/linalg/owl_base_linalg_generic.ml
```diff
@@ -191,4 +194,192 @@ let linsolve_gauss a b =
...
+let tridiag_solve_vec a b c r =
+  let n = Array.length a in
+  let n1 = Array.length b in
+  let n2 = Array.length c in
+  assert (n = n1 && n = n2);
+  if b.(0) = 0.
+  then (
+    Owl_log.error
+      "Error in tridiag_solve_vec: zero element at the beginning of diagonal vector.";
+    exit 1);
```
mseri: As a rule of thumb no library should ever call exit, raising an error here (and below) should be preferred imho. Imagine running this in a julyter notebook, you could try to recover and do something else with an error. With exit you need to restart from the beginning

ryanrhymes: Ouch! A good catch! @jzstark can you fix it?

jzstark: Thanks for pointing it out! I didn't think about this issue. I'll fix it with a new PR soon.

Issue #460, opened 2019-11-26
stochastic test failures of algodiff reverse matrix.040 [sylvester.] and discrete_lyapunov


bcdarwin:
building against nixpkgs dependencies according to https://github.com/NixOS/nixpkgs/pull/74274 .

ryanrhymes:
Calvin @tachukao is actually investigating this issue at the moment. It might be numerical stability issue because most of the time it passes on our machines.

Calvin checked `Linalg.trsyl` function it relies on, it turns out that Owl has correctly implemented trsyl whereas Scipy's seems buggy, he is preparing a pull req to scipy ;)

For this issue, the investigation is ongoing, @tachukao can comment more in case I miss anything or misunderstood anything.

tachukao:
@bcdarwin  currently we are testing the reverse mode gradients by comparing them with finite difference approximations of these gradients, which are prone to numerical errors.
I'm still unsure whether these tests are failing due to numerical errors in the finite difference approximations or from the calculations of reverse mode gradients. As @ryanrhymes mentioned, the investigation is ongoing...

mseri:
Hopefully this is fixed by #467

mseri:
@bcdarwin can you check if it fixed the issue for you?

bcdarwin:
Seems to be passing reliably now!

mseri:
I will draft a new release of owl before next week

PR #467, opened 2019-12-05
changes to algodiff tests + bug in algodiff concatenate

tachukao:
- fixed bug in the implementation of `concatenate` in Algodiff
- some changes which hopefully make the algodiff tests for `sylvester` and `discrete_lyapunov` more stable


tachukao:
To improve finite difference approximations of reverse-mode gradients, in the latest commit, I've added the option to use a fourth-order approximation. Hopefully this will help reduce random algodiff test failures referenced in #460 and other places.

mseri:
Can we make the tests quieter and show on,y the failing tests if any?

tachukao:
Good idea, I’ll give it a go

jzstark:
BTW, just wondering why the commits are not squashed...AFAIK that's kind of a common practice in PR.

mseri:
In most of the project I work on we squash only unclean histories. If the commit tree is clean as it should be and as it is in this case it is good to have a breakdown of what each commit fixes/does.

PR #470, opened 2019-12-07
Save and load Npy files

tachukao:
- Added functions to load and save npy files. These functions are implemented using [npy-ocaml](https://github.com/LaurentMazare/npy-ocaml)
- For "save" functions such as `Mat.save` and `Mat.save_txt`, use labelled argument `~out` to specify file name. Currently, to save a matrix  `mat` as a text file `mat.txt` we would do something like this
```ocaml
let () = 
  let mat = ... in
  Mat.save_txt mat "out.txt"
```
However, `mat` often comes at the end of multiple steps of calculations, it would be convenient to  "pipe" it through `Mat.save_txt` like this
``` ocaml
let () = ... |> Mat.save_txt ~out:"mat.txt"
```

### DIFF
--- a/owl-top.opam
+++ b/owl-top.opam
@@ -20,5 +20,4 @@ depends: [
   "ocaml-compiler-libs"
   "owl" {>= version}
   "owl-zoo"
-  "npy"
 ]
--- a/owl-top.opam
+++ b/owl-top.opam
@@ -20,4 +20,5 @@ depends: [
   "ocaml-compiler-libs"
   "owl" {>= version}
   "owl-zoo"
+  "npy"
 ]
--- a/owl-zoo.opam
+++ b/owl-zoo.opam
@@ -26,4 +26,5 @@ depends: [
   "dune" {>= "1.7.0"}
   "ocaml-compiler-libs"
   "owl" {>= version}
+  "npy"
 ]
--- a/owl-zoo.opam
+++ b/owl-zoo.opam
@@ -26,5 +26,4 @@ depends: [
   "dune" {>= "1.7.0"}
   "ocaml-compiler-libs"
   "owl" {>= version}
-  "npy"
 ]
--- a/src/owl/dense/owl_dense_matrix_c.mli
+++ b/src/owl/dense/owl_dense_matrix_c.mli
@@ -411,15 +411,15 @@ val of_cols : mat array -> mat
 
 val print : ?max_row:int -> ?max_col:int -> ?header:bool -> ?fmt:(elt -> string) -> mat -> unit
 
-val save : mat -> out:string -> unit
+val save : out:string -> mat -> unit
 
 val load : string -> mat
 
-val save_txt : ?sep:string -> ?append:bool -> mat -> out:string -> unit
+val save_txt : ?sep:string -> ?append:bool -> out:string -> mat -> unit
 
 val load_txt : ?sep:string -> string -> mat
 
-val save_npy : mat -> out:string -> unit
+val save_npy : out:string -> mat -> unit
 
 val load_npy : string -> mat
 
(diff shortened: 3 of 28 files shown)


tachukao:
@ghennequin

owl-top.opam
```diff
@@ -20,4 +20,5 @@ depends: [
   "ocaml-compiler-libs"
   "owl" {>= version}
   "owl-zoo"
+  "npy"
```
ryanrhymes: Do we really need this if owl already depends on npy?

owl-zoo.opam
```diff
@@ -26,4 +26,5 @@ depends: [
   "dune" {>= "1.7.0"}
   "ocaml-compiler-libs"
   "owl" {>= version}
+  "npy"
```
ryanrhymes: same here, I think owl.opam already specifies this dependency.

src/owl/dense/owl_dense_matrix_generic.ml
```diff
@@ -600,6 +600,15 @@ let load_txt ?(sep="\t") k f =
   close_in h; x
 
 
+let save_npy x ~out = Npy.write x out
+
+
+let load_npy kind file =
+  match Npy.read_copy file |> Npy.to_bigarray Bigarray.c_layout kind with
+  | Some x -> x
+  | None -> failwith Printf.(sprintf "%s: incorrect format" file)
```
jzstark: A tiny arrow match issue here, but should be no problem as we are increasingly using ocamlformat :)

owl-zoo.opam
tachukao: ar okay I wasn't sure. will remove.

src/owl/dense/owl_dense_matrix_generic.ml
```diff
@@ -565,14 +565,14 @@ let of_array k x m n =
   Owl_dense_ndarray_generic.reshape y [|m; n|]
 
 
-let save_txt ?(sep="\t") ?(append=false) x f =
+let save_txt ?(sep="\t") ?(append=false) x ~out =
```
jzstark: Agree with putting matrix as the last element. Just a matter of coding style preference, but perhaps also move the labelled arguments before `x`?  Aside from these parts, this PR looks quite good to me.

ryanrhymes: Similar comment from me, would it cause a lot of troubles if we simply move matrix as the last input argument? I'm fine with keeping out as named argument anyway.

src/owl/dense/owl_dense_matrix_c.mli
```diff
@@ -411,14 +411,17 @@ val of_cols : mat array -> mat
 
 val print : ?max_row:int -> ?max_col:int -> ?header:bool -> ?fmt:(elt -> string) -> mat -> unit
 
-val save : mat -> string -> unit
+val save : mat -> out:string -> unit
```
mseri: These are breaking changes. Are we sure we want to specify out as a labelled argument? I am not sure it is work that the breakage

src/owl/dense/owl_dense_matrix_generic.ml
mseri: I am wary of breaking changes unless they really improve something. In this case I don’t really see the benefit, matrix and output are clearly specified by the type and flipping the argument is  fast with an inline fun declaration when in a pipe

tachukao: yea the reason I kept the labelled argument at the very end is to avoid breaking changes, but it seems like `dune` enforces labels when there are labelled arguments, so perhaps the point is moot anyway.

in terms of benefit, inline function declaration definitely does the job. but it seems like we end up writing so many of these that it might justify the breaking change in my opinion...

ryanrhymes:
Looks very good after swapping the arguments, if nobody objects, I will merge the PR in the afternoon.

mseri:
We should use owl’s docker image to run a test on all the packages that depend on owl as well, so that we can help fixing breakages. Sort of like rust does with the cargo world builds right now

mseri:
I did something similar when I was at Citrix. I’ll try to design one as soon as I have time next year

ryanrhymes:
> We should use owl’s docker image to run a test on all the packages that depend on owl as well, so that we can help fixing breakages. Sort of like rust does with the cargo world builds right now

This sounds really cool!

PR #472, opened 2019-12-11
Interface files for base/dense and base/linalg

tachukao:
This is a follow-up to the last PR #471. I have added interface files for files in `src/base/linalg` and `src/base/dense`; and use these signatures in for the interface files in `src/owl/dense` and `src/owl/linalg`. 

Functions that are defined in the core interface files are the ones that have not been (fully) implemented in base. The hope is as we implement more and more functions in base, we would need less and less signatures in the core interface files.

I've created `owl_base_dense_matrix_{s,c,d,z}` files so that the structure mirrors that which we see in `src/owl/dense`. A lot remains to be done.

@ryanrhymes @jzstark I found that there were two implementations of `inv` in base and I assumed that one of them is deprecated ([link](https://github.com/tachukao/owl/blob/a75feca7778b8a51f01bb4c3839402652ddb962d/src/base/linalg/owl_base_linalg_generic.ml#L365 )). Is this right?


ryanrhymes:
> 3. One thing I'm still bit confused about the relation between `owl_base_dense_ndarray_intf.ml` and `base/types/owl_types_ndarray_basic.ml`. So my understand is that  for now if a function is updated in one file, then it is likely that the other one should also change accordingly. Is that right?

those in `base/types` are carefully crafted to provide a minimal set of functions Algodiff needs to implemented. One needs to be careful when modifying those files since it may break algodiff, dnn, cgraph ....

on the other hands, the ones Calvin made are simply for defining the interfaces implemented in Ndarray or Linalg modules. Though ideally, they shall/can also be in types folder.

src/base/dense/owl_base_dense_matrix_intf.ml
```diff
@@ -0,0 +1,20 @@
+(*
+ * OWL - OCaml Scientific and Engineering Computing
+ * Copyright (c) 2016-2019 Liang Wang <liang.wang@cl.cam.ac.uk>
+ *)
+
+
+module type Common = sig
```
ryanrhymes: Did you use `Base` in previous intf? https://github.com/owlbarn/owl/blob/0fdf627a8633e0ee26ee3c44526e3e55aa67acca/src/owl/dense/owl_dense_ndarray_intf.ml#L9

Shall we stay consistent here? Or did I miss any obvious reason for not doing so?

ryanrhymes:
For inconsistent APIs in Base.Linalg, Jianxin already had a plan for fixing them and extending to complex numbers, see this issue https://github.com/owlbarn/owl/issues/473

tachukao:
I’ve changed Base to Common in all the intf. The reason for this change was that I didn’t want to confuse Base in the interface with owl-base. I chose Common to because these signatures common to both real and complex ndarrays. But to be honest I don’t have a strong opinion about this. Happy to change it back to Base or something else. Any ideas?

ryanrhymes:
> I’ve changed Base to Common in all the intf. The reason for this change was that I didn’t want to confuse Base in the interface with owl-base. I chose the name Common because these signatures are common to both real and complex ndarrays. But to be honest I don’t have a strong opinion about this. Happy to change it back to Base or something else. Any ideas?

`Common` is fine, as long as they are used consistently in the code. I agree Base might be confusing, so let's stay with common.

src/owl/linalg/owl_linalg_intf.ml
```diff
@@ -3,18 +3,15 @@
...
-  type mat
-  type complex_mat
-  type int32_mat
+module type Common = sig
+  include Owl_base_linalg_intf.Common
+  (* NOTE: functions below have not been implemented in Base Linalg *)
 
   (** {6 Basic functions} *)
 
   val inv : mat -> mat
```
jzstark: So here we have `val inv` in Common, and in L86 we have `include Owl_base_linalg_intf.Real`, which also includes an `inv` in Real. I'm not sure if that's intended.

src/base/dense/owl_base_dense_matrix_intf.ml
tachukao: updated in the latest commit

src/base/dense/owl_base_dense_matrix_generic.ml
```diff
@@ -0,0 +1,19 @@
+(*
+ * OWL - OCaml Scientific and Engineering Computing
+ * Copyright (c) 2016-2019 Liang Wang <liang.wang@cl.cam.ac.uk>
+ *)
+
+include Owl_base_dense_ndarray_generic
```
jzstark: Logically I think perhaps we should also move matrix-specific operations such as `dot`, `trace` etc. here, but currently in the core module they are also implemented in `core/ndarray` instead of `core/matrix`, so I think it's better to leave it as some future work, not to make this quite nice PR over-complex.

src/owl/linalg/owl_linalg_intf.ml
tachukao: yea...so this is due to the fact that `inv` is only implemented for real matrices in base at the moment, but in core it's implemented for both real and complex matrices. Luckily duplicate `val` signatures doesn't seem to be a problem in ocaml...

Another option would be to move the signature of `inv` in base from `Real` to another module type called `To_be_implemented`. This module type will not be used in the core signatures. What do you think?

src/base/dense/owl_base_dense_matrix_generic.ml
tachukao: agreed!

src/owl/linalg/owl_linalg_intf.ml
jzstark: If `inv` is the only special case here, I think it's OK leave it here for now. Could you please add some comment here such a TODO to note this issue, and I'll try to update the `inv` function to cover complex input soon.

tachukao: done!

src/base/dense/owl_base_dense_ndarray.ml
```diff
@@ -23,13 +23,78 @@ module S = struct
   include Owl_base_dense_ndarray_s
   include Operator
   module Scalar = Owl_base_maths
+
+  let eye = Owl_base_dense_matrix_s.eye
```
jzstark: One last thing is that, with the signature constraint, now these extra matrix/linalg functions in ndarray could only be defined in `owl_base_dense_ndarray.ml`. I think it's actually good to be specific about which functions are aliases from other modules. Anyway, thanks for this really good PR!

tachukao: Yea I agree it’s best not to alias the functions like this. The same thing is also done in owl_dense_ndarray.ml   

This is done mainly for the benefit of algodiff. But I think there might be a better way which does not involve injecting linalg functions into ndarray

tachukao: I can try and cook up something in a different PR

src/base/dense/owl_base_dense_matrix_generic.ml
ryanrhymes: Those functions are there because because Algodiff needs them. So make sure the Algodiff will be broken. I agree logically we should keep the functions in their own module rather than duplicating the signature, the reality is more complicated though.

I tried but failed honestly, but someone may come up smarter way to reorganise the function signature so both Algodiff is happy and Ndarray signature is clean. But for now, let's keep this PR compact and avoid making things too complicatged.

ryanrhymes:
@tachukao do you still any pending commits? if not, I will merge it then.

tachukao:
None left. Sounds good!

PR #474, opened 2019-12-12
Port code to dune2

mseri:
Following what happened in eigen


mseri:
Note, the last commit should start making a difference only after the new eigen is released in opam

Issue #453, opened 2019-11-11
Overhead on matrix operations for small matrices 


pveber:
After switching from `lacaml` to `owl` on one of my projects, I observed a significant (x2) performance regression. I managed to track it down to the fact that I'm performing a lot of matrix operations on small (4x4, typically) matrices. 
So it seems that `owl` introduces a non negligible overhead for `lapack` operations. In order to document the problem and make it reproducible I wrote a tiny [benchmark](https://github.com/pveber/owl-perfs). I also noticed that vector/matrix creation via `init` function is also notably slower. 

Is there any chance to reduce that overhead?

pveber:
In order to make it easier to experiment, I created a branch in `owl` repository, see [there](/pveber/owl/tree/perf-vs-lacaml)

ryanrhymes:
This benchmark looks cool! I am (kinda) aware of this but nobody really did any benchmarking before to show how slow it is, so really appreciate this. 

Lacaml uses LAPACK whereas Owl uses LAPACKE due to c-layout, some overhead is because of this, though not very noticeable when the matrices are big. Also, Owl internally uses genarray which requires conversion between array1 and genarray in `init` function which will also introduce some overhead.

Probably need to even breakdown inside the function, to see which operation takes longer time then decide how to optimise.

ryanrhymes:
Acturally, it would be very interesting to know -- how the perfermance difference varies as a function of matrix size ... probably at some point, both lacaml and owl reach similar performance.

pveber:
Actually the conversion (flattening) happens also when multiplying matrices, and AFAIU this is partly where the difference comes from for small matrices. I pushed a new commit on the branch which defines several `partial_dot` functions, that are copies of `dot`, which go respectively until check, alloc and flatten steps (the only remaining step is then actually calling `gemm`). 

Here is the result:
```
┌───────────────────────────┬──────────┬─────────┬────────────┐
│ Name                      │ Time/Run │ mWd/Run │ Percentage │
├───────────────────────────┼──────────┼─────────┼────────────┤
│ lacaml-mat-vec-mul-4      │ 220.29ns │  22.00w │     37.10% │
│ owl-check-mat-vec-mul-4   │  50.62ns │  26.00w │      8.53% │
│ owl-alloc-mat-vec-mul-4   │ 106.83ns │  37.00w │     17.99% │
│ owl-flatten-mat-vec-mul-4 │ 345.95ns │ 107.00w │     58.26% │
[...]

PR #475, opened 2019-12-15
Remove an expensive call to flatten in `ndarray.init*`

mseri:
I hope this will mitigate https://github.com/owlbarn/owl/issues/453 (@pveber would you mind checking, I cannot run the benchmark at the moment).

Please be careful when reviewing. What was the reason of going through flatten in the first instance?

### DIFF
--- a/src/owl/dense/owl_dense_ndarray_generic.ml
+++ b/src/owl/dense/owl_dense_ndarray_generic.ml
@@ -145,26 +145,24 @@ let flatten x = reshape x [|numel x|]
 
 
 let init k d f =
-  let x = empty k d in
-  let y = array1_of_genarray (flatten x) in
-  let n = numel x in
+  let n = Array.fold_right (fun c a -> c * a) d 1 in
+  let y = Array1.create k c_layout n in
   for i = 0 to n - 1 do
     Array1.unsafe_set y i (f i)
   done;
-  x
+  reshape (genarray_of_array1 y) d
 
 
 let init_nd k d f =
-  let x = empty k d in
-  let y = array1_of_genarray (flatten x) in
-  let n = numel x in
+  let n = Array.fold_right (fun c a -> c * a) d 1 in
+  let y = Array1.create k c_layout n in
   let s = Owl_utils.calc_stride d in
   let j = Array.copy s in
   for i = 0 to n - 1 do
     Owl_utils.index_1d_nd i j s;
     Array1.unsafe_set y i (f j)
   done;
-  x
+  reshape (genarray_of_array1 y) d
 
 
 let same_shape x y = (shape x) = (shape y)
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ script: bash -ex ./.travis-docker.sh
 
 env:
   global:
-    - EIGENCPP_OPTFLAGS="-Wno-ignored-attributes -O3 -Ofast -march=native -mfpmath=sse -funroll-loops -ffast-math"
     - PACKAGE="owl"
     - PINS="owl-base:. owl-plplot:. owl-top:. owl-zoo:. owl:."
 


src/owl/dense/owl_dense_ndarray_generic.ml
```diff
@@ -145,26 +145,24 @@ let flatten x = reshape x [|numel x|]
...
 
 
 let init_nd k d f =
-  let x = empty k d in
-  let y = array1_of_genarray (flatten x) in
-  let n = numel x in
+  let n = Array.fold_right (fun c a -> c * a) d 1 in
+  let y = Array1.create k c_layout n in
   let s = Owl_utils.calc_stride d in
   let j = Array.copy s in
```
mseri: @ryanrhymes why do you need this copy?

ryanrhymes: if user function `f` modifies `s`, then the algorithm breaks; by passing a copy, we are safe no matter what `f` does to `j`.

mseri: Ah Yes, good point!

mseri:
I have found a moment to run the benchmarks.

With this patch:

```
❯ dune exec ./owl_perfs.exe mat-mat-mul-small
Estimated testing time 20s (2 benchmarks x 10s). Change using '-quota'.
┌──────────────────────┬──────────┬─────────┬────────────┐
│ Name                 │ Time/Run │ mWd/Run │ Percentage │
├──────────────────────┼──────────┼─────────┼────────────┤
│ lacaml-mat-mat-mul-4 │ 352.58ns │  18.00w │     51.70% │
│ owl-mat-mat-mul-4    │ 681.94ns │ 148.00w │    100.00% │
└──────────────────────┴──────────┴─────────┴────────────┘


❯ dune exec ./owl_perfs.exe vec-init-small
Estimated testing time 20s (2 benchmarks x 10s). Change using '-quota'.
┌───────────────────┬──────────┬─────────┬────────────┐
│ Name              │ Time/Run │ mWd/Run │ Percentage │
├───────────────────┼──────────┼─────────┼────────────┤
│ lacaml-vec-init-4 │ 133.54ns │   9.00w │     43.99% │
│ owl-vec-init-4    │ 303.57ns │  28.00w │    100.00% │
└───────────────────┴──────────┴─────────┴────────────┘
```

Without this patch:
```
❯ dune exec ./owl_perfs.exe vec-init-small
Estimated testing time 20s (2 benchmarks x 10s). Change using '-quota'.
┌───────────────────┬──────────┬─────────┬────────────┐
│ Name              │ Time/Run │ mWd/Run │ Percentage │
├───────────────────┼──────────┼─────────┼────────────┤
│ lacaml-vec-init-4 │ 133.63ns │   9.00w │     38.13% │
│ owl-vec-init-4    │ 350.44ns │  32.00w │    100.00% │
└───────────────────┴──────────┴─────────┴────────────┘

~/code/tests/owl-perfs master 22s
❯ dune exec ./owl_perfs.exe vec-init-large
Estimated testing time 20s (2 benchmarks x 10s). Change using '-quota'.
┌─────────────────────┬────────────┬─────────┬────────────┐
│ Name                │   Time/Run │ mWd/Run │ Percentage │
├─────────────────────┼────────────┼─────────┼────────────┤
│ lacaml-vec-init-100 │   399.91ns │   9.00w │     31.22% │
│ owl-vec-init-100    │ 1_280.81ns │  32.00w │    100.00% │
└─────────────────────┴────────────┴─────────┴────────────┘
```

mseri:
On https://discuss.ocaml.org/t/working-with-a-huge-data-chunks/3955/10 I have read
```
Creating a view, e.g., via reshape or sub is usually slow, because it is a custom value and has to be created in the major heap, thus every creating of a bigarray triggers the minor collection. We were addressing those issues by preparing all the views that we need in advance (or having a pool of views). [...]
```
Maybe this is really the culprit and we need to find an alternative mechanism.

pveber:
Thanks for this attempt @mseri! I was wondering: the underlying type for matrices could be a sum type, that is a `Mat.t` could be:
```ocaml
type t =
  | Gen of (...) Genarray.t
  | Mat2 of (...) Array2.t
  | ...
```
When using only lapack operations, we could then avoid all overhead while keeping the API unchanged (and simple).

On a completely orthogonal note, I have found *extremely* useful to [type vectors and matrices operations in a rather strict way](https://github.com/biocaml/phylogenetics/blob/b7ca01042bb53ba09158307063f8c0237faca9f6/lib/linear_algebra.mli), it showed me lots of bugs compared to using `owl` API directly.

mseri:
> On a completely orthogonal note, I have found _extremely_ useful to [type vectors and matrices operations in a rather strict way](https://github.com/biocaml/phylogenetics/blob/b7ca01042bb53ba09158307063f8c0237faca9f6/lib/linear_algebra.mli), it showed me lots of bugs compared to using `owl` API directly.

I think it is worth opening an issue about this (See now https://github.com/owlbarn/owl/issues/480). I have also been coming back and forth to this issueFrom time to time.

mseri:
> the underlying type for matrices could be a sum type, that is a `Mat.t` could be:
> 
> ```ocaml
> type t =
>   | Gen of (...) Genarray.t
>   | Mat2 of (...) Array2.t
>   | ...
> ```
> 
> When using only lapack operations, we could then avoid all overhead while keeping the API unchanged (and simple).

This is an interesting idea. It may be worth to try and give it a go at a certain stage. We are in principle free to rebind operations in the modules with more efficient ones. Lots of this code could actually be auto-generated with a suitable ppx

nilsbecker:
An orthogonal dimension where typing could in principle help improve numerical efficiency is to include matrix structure in the types. LAPACK/BLAS have specialized routines for diagonalizing/decomposing e.g. symmetric or hermitian matrices. some matrix functions are known to preserve these properties, which could also be propagated in their (GAD)type in principle. e.g. eigenvalues of hermitian matrices are real. i think there is/was a ruby numerical library that did something like this, but could not confirm this now when googling Numo.

of course, a somewhat overlapping issue is that matrices with a certain structure can also be stored in various optimal ways (e.g. bands, only one half of symmetric matrices, etc). afaik it's an open question whether a system that tries to automatically dispatch matrices to/from different sparse storage formats according to their structure, can be useful. it might be doing more harm than good by making memory usage unpredictable.

PR #476, opened 2019-12-17
use a separate module for algodiff instead of ndarray directly

tachukao:
- created `owl_algodiff_primal_ops.ml` to host all the ndarray, linalg, and matrix functions that are used for Algodiff in one place. The main advantage here is that we don't need to inject all the algodiff-required operations in to the ndarray module as was done previously (see #472). 

- refactored `owl_algodiff_build_ops.ml`. previously I was exposing functions like `op_siso`, but there really is no need for this. I've put all the relevant functions close to each other so the code is a bit more readable, but it's still lacking in documentation.

Ideally, perhaps in a separate PR, we would have a submodule called `Arr` in `owl_algodiff_primal_ops.D`. This module includes all the Ndarray functions. This would involve changing quite a lot of  the code . As a concrete example, we would  need to change `A.float_to_elt` to `A.Arr.float_to_elt` here

https://github.com/tachukao/owl/blob/b52264635cc9d3019c610c272e5baae4f8f3fa0e/src/base/compute/owl_computation_operator.ml#L67 

If we make this change,  I think it would also make sense to rename the module `A` to `PrimalOps` in `Algodiff`. Conceptually this would make things a lot more consistent. `PrimalOps` would include submodules `Arr`, `Mat`, and `Linalg`, which are the  _primal_ operations that are used for algodiff.


mseri:
This looks like a nice cleanup

ryanrhymes:
How about creating a algodiff directory in owl like that in base, rather than storing algodiff related files in optimise directory?

mseri:
Should we also consider, in a separate PR, to drop all the owl_ and to define sublibraries for the various su packages? I’d be happy to give it a try in a week or so

mseri:
@jzstark there is only the primal ops module to plumb, or I missed something? I actually like that primal ops include the various submodules separately now

mseri:
@tachukao are the examples still working? If not we should update them

jzstark:
Hi Marcello, as I said, I think practically this is a really nice way to separate different modules. My worry is just about the design issue. In Owl the ndarray is meant as the only foundamental data structure to support the layers above, e.g. CGraph, AD, symbolic, etc. There is a consistency of interfaces. Now that we have two sets of such data structure, that might lead to a confusion in using.  

That being said, I think this PR has more pros than cons (perhaps the naming of files could be updated? "algdiff_primal_ops" sounds a bit confusing to me...), and I'm happy to accept it. 

Btw, I'm not clear about the motivation of dropping the "owl_" headers of all the files.

tachukao:
> How about creating a algodiff directory in owl like that in base, rather than storing algodiff related files in optimise directory?

done! @ryanrhymes 

> Thanks Calvin! This PR separates out the Mat/Linalg functions in Ndarray, which I think is really nice. My only concern is about the idea of low level ndarrys: so now instead of `owl_ndarray` and `owl_base_ndarray`, the users also have `owl_base_algodiff_primal_ops` and `owl_base_algodiff_primal_ops` to plug in functors such as in CGraph. Practically I think this new mechanism works quite well, but I'm not quite sure if that's a sound design choice.

@jzstark `owl_algodiff_primal_ops` is the only interface that satisfies `Owl_types_ndarray_algodiff.Sig` now. `Owl_dense_ndarray_s.ml` no longer works.

tachukao:
> @tachukao are the examples still working? If not we should update them

@mseri as far as I'm aware, the examples never actually make and Algodiff module. They simply use the _made_ `Algodiff.D` or `Algodiff.S` modules which work fine. But maybe @ryanrhymes can confirm this?

tachukao:
> That being said, I think this PR has more pros than cons (perhaps the naming of files could be updated? "algdiff_primal_ops" sounds a bit confusing to me...), and I'm happy to accept it.

@jzstark @mseri @ryanrhymes any suggestions for naming? open to new ideas...not entirely happy with the name. Briefly considered `owl_algodiff_pop`, where `pops` stands for `primal ops`...

mseri:
> I think practically this is a really nice way to separate different modules. My worry is just about the design issue. In Owl the ndarray is meant as the only foundamental data structure to support the layers above, e.g. CGraph, AD, symbolic, etc. There is a consistency of interfaces. Now that we have two sets of such data structure, that might lead to a confusion in using.

> That being said, I think this PR has more pros than cons (perhaps the naming of files could be updated? "algdiff_primal_ops" sounds a bit confusing to me...), and I'm happy to accept it.

Sure. I wanted to understand better your concerns. I think this is the time and place to try and figure out the correct design. We should not rush the merge  For me personally it would be useful to see a couple of nontrivial examples before and after this PR

mseri:
> Btw, I'm not clear about the motivation of dropping the "owl_" headers of all the files.

Sorry, I mean that we are doing namespacing manually instead of letting dune deal with it. But it is OT here now
I’ll move the discussion and explain myself properly in a new issue in the next few days. Let’s not clutter this PR

ryanrhymes:
> > That being said, I think this PR has more pros than cons (perhaps the naming of files could be updated? "algdiff_primal_ops" sounds a bit confusing to me...), and I'm happy to accept it.
> 
> @jzstark @mseri @ryanrhymes any suggestions for naming? open to new ideas...not entirely happy with the name. Briefly considered `owl_algodiff_pops`, where `pops` stands for `primal ops`...

Owl engine in Owl Symbolic also needs a similar module, we can also name them consistently e.g. like `owl_ndarray_symbolic`, `owl_ndarray_algodiff`, so we know they are stub modules to provide matching module signature. But we can discuss this naming thing later in another PR.

mseri:
Why not waiting for the renaming before merging?

Issue #484, opened 2020-01-06
Graph.inputs not behaving as expected


mreppen:
The behavior of Graph.inputs is not as expected.
```
open Owl
open Neural.S
open Neural.S.Graph
open Neural.S.Algodiff
let length = 5
let batch_size = 3
let make_network () =
  inputs [| [| length |]; [| length |] |]
  |> mul
  |> get_network
let nn = make_network ()
let data = Dense.Ndarray.S.sequential [|batch_size; length |]
let result = Graph.model nn data
```
duplicates the input:
```
val data : Owl.Dense.Ndarray.S.arr =
  
   C0 C1 C2 C3 C4 
R0  0  1  2  3  4 
R1  5  6  7  8  9 
R2 10 11 12 13 14 

val result : Owl_dense_ndarray.S.arr =
  
    C0  C1  C2  C3  C4 
R0   0   1   4   9  16 
R1  25  36  49  64  81 
R2 100 121 144 169 196
```

Instead, I expect it to read multiple different arrays.

With `inputs [| [| length_one |]; [| length_two |] |]` and `length_one <> length_two` I have not been able to make it accept any data. For instance, I would expect to do something in the style of
```
open Owl
open Neural.S
open Neural.S.Graph
open Neural.S.Algodiff
let length = 5
let batch_size = 3
let make_network () =
  inputs [| [| length |]; [| 1 |]|]
  |> concatenate 1
  |> get_network
let nn = make_network ()
let data = Dense.Ndarray.S.sequential [|batch_size; Stdlib.(length + 1) |]
let result = Graph.model nn data;;
Exception: Assert_failure ("src/base/neural/owl_neural_neuron.ml", 113, 10).
```

mreppen:
In hindsight, it could be that I misunderstood something. I will have another look when I find the time.

PR #493, opened 2020-01-30
Make Graph.inputs give unique names to inputs

mreppen:
I noticed that `Graph.inputs` gives all inputs the name `"input_0"`. I assume nodes are expected to have unique names, so I moved the call `add_node` to right after the node creation. Does this make sense?

### DIFF
--- a/src/base/neural/owl_neural_graph.ml
+++ b/src/base/neural/owl_neural_graph.ml
@@ -315,10 +315,13 @@ module Make (Neuron : Owl_neural_neuron_sig.Sig) = struct
     let neurons = Array.map (fun s -> Input (Input.create s)) input_shapes in
     let nn = make_network 0 [||] [||] in
     let ns =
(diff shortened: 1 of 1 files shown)


ryanrhymes:
@pvdhove I remember you added this function while working on MRCNN. Do you have any specific reason for assigning the same name? Otherwise, the PR looks fine by me. @jzstark opinions?

jzstark:
This change looks good to me. I think each input should be unique by nature. I think this is also related to issue #484 you have mentioned. This operation is introduced in building the application MRCNN. As long as this application is not broken, I'm totally OK with this PR.

pvdhove:
No, it was not my intention to give the same name to the inputs (I had always given them explicit names so I had never noticed). The change is clearly OK and MRCNN still works fine!

jzstark:
Cool! Thanks a lot @pveber!

ryanrhymes:
Thank you Pierre for clarifying this.

PR #495, opened 2020-02-10
Broadcast bug for higher order derivatives

tachukao:
Currently,  when the forward pass involves broadcasting operations, we use `_shrink` in `owl_algodiff_reverse.ml` to make sure the adjoints have the right shape.
Because `_shrink` is not differentiable, we run into reshape/slicing errors when we try and differentiate through it to calculate higher order derivatives e.g.`hessian`.

This is not a problem when we are dealing with first-order gradients because we do not need to differentiate through `_shrink`.

In this PR, I fix this problem by removing the `_shrink` operation from `owl_algodiff_reverse.ml` and introduce a similar, but differentiable, function `_squeeze_broadcast` in `owl_algodiff_ops.ml`. This function is used to make sure that broadcast operations always return adjoints that have the right shape.

Here is a simple script that fails and is fixed in this PR

```ocaml
open Owl
module AD = Algodiff.D
let f x =
   let y = AD.Maths.get_slice [[]; [0]] x in
   let z = AD.Maths.get_slice [[]; [1;4]] x in
   AD.Maths.(sum' (exp (z+y)))
let () =
   let h = AD.(grad (grad f)) in
   let x = AD.Mat.gaussian 1 5 in
   h x |> ignore 
```

I have also changed the `shape` function in algodiff so that it returns an empty `int array` for a float.

### DIFF
--- a/src/base/algodiff/owl_algodiff_ops.ml
+++ b/src/base/algodiff/owl_algodiff_ops.ml
@@ -14,34 +14,34 @@ module Make (Core : Owl_algodiff_core_sig.Sig) = struct
   open Builder
 
   module Maths = struct
-    (* squeeze x so that it is same size as y *)
-    let rec squeeze_broadcast x y =
-      if is_float y
-      then if is_float x then x else sum' x
-      else if is_float x
-      then failwith "x cannot be float while y is not"
+    (* squeeze x so that it has shape s *)
+    let rec squeeze_broadcast x s =
+      let sx = shape x in
+      let lx = Array.length sx in
+      let ls = Array.length s in
+      if sx = s
+      then x
+      else if lx < ls
+      then failwith "x should have higher dimensions than that specified by s"
+      else if ls = 0
+      then sum' x
       else (
-        let sx = shape x in
-        let s = shape y in
-        if sx = s
-        then x
-        else (
-          let _, idxs =
-            Array.fold_left
-              (fun (k, accu) sx ->
-                if s.(k) = sx
-                then succ k, accu
-                else if s.(k) = 1
-                then succ k, k :: accu
-                else
-                  failwith
-                    Printf.(
-                      sprintf "there ought to be a broadcasting error %i, %i\n%!" s.(k) sx))
-              (0, [])
-              sx
-          in
-          let idxs = Array.of_list idxs in
-          sum_reduce ~axis:idxs x))
+        let _, idxs =
+          Array.fold_left
+            (fun (k, accu) sx ->
+              if s.(k) = sx
+              then succ k, accu
+              else if s.(k) = 1
+              then succ k, k :: accu
+              else
+                failwith
+                  Printf.(
+                    sprintf "there ought to be a broadcasting error %i, %i\n%!" s.(k) sx))
+            (0, [])
+            sx
+        in
+        let idxs = Array.of_list idxs in
+        sum_reduce ~axis:idxs x)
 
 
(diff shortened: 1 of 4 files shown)


src/base/algodiff/owl_algodiff_ops.ml
```diff
@@ -14,8 +14,38 @@ module Make (Core : Owl_algodiff_core_sig.Sig) = struct
...
   module Maths = struct
+    (* squeeze x so that it has shape s *)
+    let rec _squeeze_broadcast x s =
+      let sx = shape x in
+      let lx = Array.length sx in
+      let ls = Array.length s in
+      if sx = s
+      then x
+      else if lx < ls
+      then failwith "x should have higher dimensions than that specified by s"
```
mseri: I would be more explicit, something like
```suggestion
      then failwith Printf.(printf "_squeeze_broadcase: x must have dimension greater than %i, instead has dimension %i" ls lx)
```

src/base/algodiff/owl_algodiff_ops.ml
```diff
@@ -14,8 +14,38 @@ module Make (Core : Owl_algodiff_core_sig.Sig) = struct
...
+          Array.fold_left
+            (fun (k, accu) sx ->
+              if s.(k) = sx
+              then succ k, accu
+              else if s.(k) = 1
+              then succ k, k :: accu
+              else
+                failwith
+                  Printf.(
+                    sprintf "there ought to be a broadcasting error %i, %i\n%!" s.(k) sx))
```
mseri: What about
```suggestion
                    sprintf "_squeeze_broadcast: unkonwn broadcasting error %i, %i\n%!" s.(k) sx))
```

mseri:
LGTM, some minor comments on the failures to try and ease debugging.
I am getting warier about broadcasting, but I guess it is a necessary evil from the moment

src/base/algodiff/owl_algodiff_ops.ml
```diff
@@ -14,8 +14,40 @@ module Make (Core : Owl_algodiff_core_sig.Sig) = struct
...
+    (* squeeze x so that it has shape s *)
+    let rec _squeeze_broadcast x s =
+      let sx = shape x in
+      let lx = Array.length sx in
+      let ls = Array.length s in
+      if sx = s
+      then x
+      else if lx < ls
+      then failwith Printf.(
+          sprintf "_squeeze_broadcase: x must have dimension greater than %i, instead has dimension %i" ls lx
```
mseri:
```suggestion
          sprintf "_squeeze_broadcast: x must have dimension greater than %i, instead has dimension %i" ls lx
```
Sorry I made a small typo :P

tachukao:
@mseri I don't think broadcasting is going to be a big problem here with algodiff. it's rather unlikely to run into these failures, unless one builds their own algodiff operations erroneously that uses broadcast...

ryanrhymes:
Good catch :) I am aware of this issue when I coded `_shrink` function. It is really great that you fixed it!

tachukao:
@jzstark `_shrink` was applied to all operations, but only really mattered for operations that use broadcast.  
I think the only operations that use broadcast are `add`, `sub`, `mul`, `div` and `pow` at the moment, but I might be wrong. If you can think of others, we will need to change those as well.

ryanrhymes:
@tachukao @jzstark here is the list https://ocaml.xyz/book/slicing.html#scrollNav-7-3

Issue #496, opened 2020-02-10
draw samples from a poisson distribution with Ndarray


tachukao:
It would be great to have a function in Ndarray that allows us to draw samples from a poisson distribution much like `Arr.gaussian` allows us to draw samples from a normal distribution.

For example, in the `Arr` module we might have
```ocaml
val poisson : mu:[< `arr of Arr.arr | `flt of float] -> int array -> Arr.arr
```
where `mu` is the rate of the poisson distribution we draw from; it can either be a `float` or and `Arr.arr`. If it is an `Arr.arr` it must be broadcastable over the requested shape as specified by the last argument `int array`.

PR #498, opened 2020-02-10
Draw poisson RVs in Ndarray and Mat modules

tachukao:
I added a function to draw samples from a poisson distribution with rate `mu`
```ocaml
val poisson : mu:float -> int array -> Arr.arr
```
Here, I have not implemented the option to provide a `mu` of type `Arr.arr`, as suggested in #496.  

Perhaps, as future extensions,  for each the random variable functions `gaussian`, `uniform` and `poisson`, we could add the variants:

```ocaml
val gaussian_rvs : mu:Arr.arr -> sigma:Arr.arr -> int array -> Arr.arr
val uniform_rvs : a:Arr.arr -> b:Arr.arr -> int array -> Arr.arr
val poisson_rvs: mu:Arr.arr -> int array -> Arr.arr
```

where parameters `mu`, `sigma`, `a`, and `b` are broadcasted to the requested shape.

For `gaussian_rvs`, this could already easily implemented:

```ocaml
let gaussian_rvs ~mu ~sigma dims =
     mu + (sigma * (gaussian dims))
```

### DIFF
--- a/src/owl/core/owl_ndarray_maths.ml
+++ b/src/owl/core/owl_ndarray_maths.ml
@@ -7044,6 +7044,28 @@ let _owl_exponential : type a b. (a, b) kind -> (a, b) owl_arr_op13 = function
   | _       -> failwith "_owl_exponential: unsupported operation"
 
 
+external owl_float32_poisson
+  :  int
+  -> ('a, 'b) owl_arr
+  -> float
+  -> int
+  -> unit
+  = "float32_poisson"
+
+external owl_float64_poisson
+  :  int
+  -> ('a, 'b) owl_arr
+  -> float
+  -> int
+  -> unit
+  = "float64_poisson"
+
+let _owl_poisson : type a b. (a, b) kind -> (a, b) owl_arr_op12 = function
+  | Float32 -> owl_float32_poisson
+  | Float64 -> owl_float64_poisson
+  | _       -> failwith "_owl_poisson: unsupported operation"
+
+
 external owl_float32_diff
   :  int
   -> int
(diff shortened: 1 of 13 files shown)


ryanrhymes:
fyi, if you need more flexible parameterisation, you can consider https://github.com/owlbarn/owl/blob/master/src/owl/ppl/owl_distribution.mli

tachukao:
Oh cool! this looks super useful. Is the `Owl_distribution` module readily _made_ anywhere? 

I'll try and add a `Poisson` module to this in the coming days. This looks great

ryanrhymes:
This module was orignally designed to serve PPL, but PPL progress was stalled. But this year, there are two Part II students working on PPL atop of owl, let's see what they will bring to us :)

The module interface borrows the design of corresponding module in TensorFlow. It is not used by any other modules atm, actually few people know this module.

ryanrhymes:
feel free to tinker it and PR :)

tachukao:
I think for general use, the functions in `Owl_distribution_generic` are already very useful, and could in principle be directly included in the `Ndarray` modules? what do you think?

ryanrhymes:
might make ndarray interface very heavy, but let see what others @mseri @jzstark think ...

mseri:
I agree with @ryanrhymes, it is better to have easy to find specific modules imho than massive ones. We also benefit from link time cleanups from the compiler removing all the unused modules from the produced artifacts

tachukao:
yea actually I agree. it's probably best kept separate. Plus, `Owl_distribution` seems to depend on the dense `Ndarray` module already

jzstark:
Agree. It's useful to add some standard distribution such as gaussian/poisson/uniform in ndarray  module, but perhaps it goes too far to move all of them into ndarray. Adding parameters to these distributions sounds a good idea. Also, perhaps we should consider adding it in the base library later if a poisson distribution is indeed frequently used.

PR #499, opened 2020-02-11
Added poisson_rvs to Owl_distribution

tachukao:
Following the last PR #498, I've added `poisson_rvs` to `Owl_distribution_generic` and `Owl_distribution` as a module.

I've also replaced the `loggamma` function with the cephes implementation `lgam` (this was on the TODO).

On a side note, I find some of the documentation in `owl_ndarray_maths_map.h` slightly confusing. I'm still not entirely sure why `exponential_rvs` uses `FUN24` and `gaussian_rvs` uses `FUN25`. I'm using `FUN24` to define `poisson_rvs`, as like `exponential_rvs` it only has one parameter. It would be nice to have a more detailed description of these functions and maybe more informative names at some point? Just a thought.

### DIFF
--- a/src/base/types/owl_types_stats_dist.ml
+++ b/src/base/types/owl_types_stats_dist.ml
@@ -60,6 +60,8 @@ module type Sig = sig
 
   val exponential_isf : lambda:arr -> arr -> arr
 
+  val poisson_rvs : mu:arr -> n:int -> arr
+
   val gamma_rvs : shape:arr -> scale:arr -> n:int -> arr
 
   val gamma_pdf : shape:arr -> scale:arr -> arr -> arr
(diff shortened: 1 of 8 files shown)


mseri:
> It would be nice to have a more detailed description of these functions and maybe more informative names at some point

I particularly agree with this comment. Touching the C interface is very non-straightforward and would be good to have a large section, maybe in `CONTRIBUTING.md` that explains what is going on and how things work.

mseri:
From what I can understand, LGTM

ryanrhymes:
There will be one chapter dedicated to the underlying c core in the book, to help others in understandibng what happens behind the scene so they can modify the code.

PR #501, opened 2020-02-21
Fix bug in reverse mode gradients of aiso operations and pow

tachukao:
In reverse mode, the gradients should be calculated using the primal of the inputs (being differentiated) as opposed to the inputs themselves. Fixed this bug for `pow` and `aiso` operations such as `sylvester` and `care`


jzstark:
Thanks for this quick fix!  It works on my example of calculating jacobian matrix of function that contains `pow`.

PR #503, opened 2020-02-24
Fix bug in _squeeze_broadcast

tachukao:
This PR fixes a bug in the function `_squeeze_broadcast` in `owl_algodiff_ops.ml`. 
This function is used to ensure that the dimensionality of the adjoint is compatible with that of the input in the forward pass. Currently `_squeeze_broadcast` does not accommodate valid broadcasting operations that involve inputs with different dimensions. For example, it would give a broadcasting error for the following operation
```ocaml
let x = a + b
```
where `a` has dimensions `[|2; 3; 4|]` and `b` has dimensions `[|3; 4|]`. 
As a result of this bug, `mnist_cnn.ml` and `lazy_mnist.ml` in the examples were broken.
This PR fixes this issue.

### DIFF
--- a/src/base/algodiff/owl_algodiff_ops.ml
+++ b/src/base/algodiff/owl_algodiff_ops.ml
@@ -523,26 +523,31 @@ module Make (Core : Owl_algodiff_core_sig.Sig) = struct
 
     and _sum =
       lazy
-        (fun ~axis ->
+        (fun ?axis ->
           build_siso
             (module struct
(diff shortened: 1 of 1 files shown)


jzstark:
Cool! Thanks for the quick fix. The `mnist_cnn.ml` and `lazy_mnist.ml` example work fine now, and `make test` all pass.  Also, the example that led to the fix in #501 still works fine.

Issue #505, opened 2020-02-26
Feature request: Ndarray stack


tachukao:
It would be useful to have a function `stack` in the Ndarray module similar to the `stack` function in numpy:

https://docs.scipy.org/doc/numpy/reference/generated/numpy.stack.html#numpy.stack

To achieve the same functionality right now we have to first expand the dimensions of the ndarrays and concatenate along that axis. The proposed function would have type signature

```ocaml
val stack : ?axis:int -> arr array -> arr
```

A naive implementation would be

```ocaml
let stack ?(axis = 0) xs =
      let shp = Arr.shape xs.(0) in
      let ndim = Array.length shp + 1 in
      let axis = Owl_utils.adjust_index axis ndim in
      let new_shp = Array.init ndim (fun i ->
             if i < axis then shp.(i)
             else if i = axis then 1
             else  shp.(i-1)) in
      let y = Array.map (fun x ->
                   let shp' = Arr.shape x in
                   if shp' <> shp then failwith "stack: ndarrays in [xs] must all have the same shape";
                   Arr.reshape x new_shp) xs in
       Arr.concatenate ~axis y
```

ryanrhymes:
The naive impl looks fine by me, for large ndarray, most of the time will be spent in concatenate function anyway.

PR #506, opened 2020-02-27
added stack function

tachukao:
added `stack` function to Ndarray modules as discussed in #505


mseri:
We have added lots lately but no new test. Should we add some test cases?

ryanrhymes:
Agree, unit tests have been very useful in owl development. It is better to keep sufficient coverage of the functions in owl.

tachukao:
added unit test for `stack` in the latest commits

Issue #489, opened 2020-01-21
Owl will not be compatible with the next version of ctypes


mseri:
See https://github.com/ocaml/opam-repository/pull/15708

yallop:
I hadn't spotted that this change would affect `owl`.

The goal of the change in `ctypes` (https://github.com/ocamllabs/ocaml-ctypes/commit/a900b59ef487d0c0f5feca5b607ef252d19af683) is to track the allocation status of some objects in the types of the OCaml values that reference them.  The change is essentially internal, but it does change the interface between the code generated by `ctypes` and the `ctypes` library (i.e. the [`Cstubs_internals`](https://github.com/ocamllabs/ocaml-ctypes/blob/a900b59ef487d0c0f5feca5b607ef252d19af683/src/ctypes/cstubs_internals.mli) interface).  Since `owl` generates code directly rather than using the `ctypes` generator, it's also affected.

The main change should be to replace
```ocaml
... cptr p ...
```

with
```ocaml
let Cpointer x = p in ... x ...
```

i.e. to use a `let` binding to unpack pointers rather than the `cptr` function.  I'm happy to help with getting `owl` working with the new interface, if it would be useful.

If this is going to cause major problems, please let me know and we can reconsider.

mseri:
Thanks! I think it will be fine. In case we hit some rough edge we will ping for help

tachukao:
we can probably close this now?!

PR #508, opened 2020-02-28
Owl now compatible with latest version of Ctypes

tachukao:
This PR makes Owl compatible with the latest version of Ctypes see #489 

Changes are made to the lapacke and cblas parser and generated files as well as `owl_maths_special.ml`.

### DIFF
--- a/owl.opam
+++ b/owl.opam
@@ -27,7 +27,7 @@ depends: [
   "base" {build}
   "base-bigarray"
   "conf-openblas" {>= "0.2.0"}
-  "ctypes" {< "0.17.0"}
+  "ctypes"
   "dune" {>= "2.0.0"}
   "dune-configurator"
   "eigen" {>= "0.1.0"}
--- a/owl.opam
+++ b/owl.opam
@@ -27,7 +27,7 @@ depends: [
   "base" {build}
   "base-bigarray"
   "conf-openblas" {>= "0.2.0"}
-  "ctypes"
+  "ctypes" {>- "0.16.0"}
   "dune" {>= "2.0.0"}
   "dune-configurator"
   "eigen" {>= "0.1.0"}
--- a/owl.opam
+++ b/owl.opam
@@ -27,7 +27,7 @@ depends: [
   "base" {build}
   "base-bigarray"
   "conf-openblas" {>= "0.2.0"}
-  "ctypes" {>- "0.16.0"}
+  "ctypes" {>= "0.16.0"}
   "dune" {>= "2.0.0"}
   "dune-configurator"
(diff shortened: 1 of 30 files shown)


owl.opam
```diff
@@ -27,7 +27,7 @@ depends: [
   "base" {build}
   "base-bigarray"
   "conf-openblas" {>= "0.2.0"}
-  "ctypes" {< "0.17.0"}
+  "ctypes"
```
mseri: Doesn’t this require the latest ctypes?
```suggestion
  "ctypes" { >= "0.17.0"}
```

mseri:
Do youneed to run ocamlformat on the generated files? I am surprised that the diff is so large

owl.opam
tachukao: these changes seem to work with older versions of ctypes as well. I tested it on 0.16.0

tachukao:
> Do youneed to run ocamlformat on the generated files? I am surprised that the diff is so large

hmmm I didn't actually apply ocamlformat, but maybe that's the problem...

owl.opam
mseri: Let’s use 0.16.0 as a bound then. We can alwys relax it if really needed, but it will reduce testing burden to us and for the opam releases

mseri:
> > Do youneed to run ocamlformat on the generated files? I am surprised that the diff is so large
> 
> hmmm I didn't actually apply ocamlformat, but maybe that's the problem...

Sorry, that’s exactly what I meant, if you needed to ocamlformat them to reduce the diff

tachukao:
Indeed, after applying ocamlformat the diffs are smaller

owl.opam
```diff
@@ -27,7 +27,7 @@ depends: [
   "base" {build}
   "base-bigarray"
   "conf-openblas" {>= "0.2.0"}
-  "ctypes"
+  "ctypes" {>- "0.16.0"}
```
mseri:
```suggestion
  "ctypes" {>= "0.16.0"}
```

yallop:
If you haven't already, it'd be worth testing this against ctypes 0.17.1 (which makes some [small changes](https://github.com/ocamllabs/ocaml-ctypes/pull/629) related to the internal-interface changes in 0.17.0).

mseri:
Indeed, thanks! @tachukao can you pin the lates ctypes and check?

tachukao:
I can't find ctypes 0.17.1 on opam yet, but I tried pinning to the dev version on github, and it seems to work.

mseri:
Great! Thanks @tachukao and @yallop

PR #515, opened 2020-04-06
Graph subnetwork: Multiple outputs

mreppen:
Hi, I finally got around to adding this small change. This was discussed in https://github.com/owlbarn/owl/pull/491 I personally don't need it, but for the sake of completeness, here it is.

Example:
```
let nn = Graph.(
  let inp = input [| 2 |] in
  let x1 = lambda ~name:"x1" ~out_shape:[|1|] (fun x -> Algodiff.Maths.get_slice [[]; [0]] x) inp in
  let x2 = lambda ~name:"x2" ~out_shape:[|1|] (fun x -> Algodiff.Maths.get_slice [[]; [1]] x) inp in
  let f1 = fully_connected ~name:"f1" 1 x1 in
  let f2 = fully_connected ~name:"f2" 1 x2 in
  let sum = add ~name:"sum" [| f1; f2 |] in
  sum |> get_network);;
Graph.get_subnetwork ~make_inputs:[|"x1"; "x2"|] nn [| "f1"; "f2" |];;
```

Two comments:
 - I also added a flag for whether to copy the weights.
 - I did change the signature (likely not many are using this yet, so hopefully not a problem). This also allowed me to change the default copy behavior to `true`. If this is undesirable, d684020389524e5b35fbe62e87160a4d09e51b26 is a version without and I can PR that instead.

PS. I have myself found this subnetwork feature really useful :)

### DIFF
--- a/src/base/neural/owl_neural_graph_sig.ml
+++ b/src/base/neural/owl_neural_graph_sig.ml
@@ -673,18 +673,16 @@ Load the weights from a file of the given name. Note that the weights and the
 name of their associated neurons are saved as key-value pairs in a hash table.
   *)
 
-  val get_subnetwork : ?copy:bool -> ?make_inputs:string array -> node -> network
+  val get_subnetwork : ?copy:bool -> ?make_inputs:string array -> network -> string array -> network
   (**
-   Constructs a subnetwork of nodes on which ``node`` depends, replacing
-   nodes with names in ``make_inputs`` with input nodes. If ``make_inputs``
-   is empty or unspecified, the original input nodes are used.  ``copy`` is
-   a bool controlling whether the weights of the new network are copies or
-   references to those of the old network (defaults to ``false``).
-   *)
+   Constructs a subnetwork of nodes on which ``output_names`` depend,
+   replacing nodes with names in ``make_inputs`` with input nodes.
 
-  val get_subnetwork_array : ?copy:bool -> ?make_inputs:string array -> network -> string array -> network
-  (**
-   Like ``get_subnetwork``, but allows for multiple output nodes in the subnetwork.
+   Arguments:
+     ``copy``: Whether to copy or reference the original node weights. Defaults to true.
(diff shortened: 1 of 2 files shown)


src/base/neural/owl_neural_graph_sig.ml
```diff
@@ -673,14 +673,16 @@ Load the weights from a file of the given name. Note that the weights and the
...
-  (**
-   Constructs a subnetwork of nodes on which ``node`` depends, replacing
-   nodes with names in ``make_inputs`` with input nodes. If ``make_inputs``
-   is empty or unspecified, the original input nodes are used.
-   
-   Note: the weights in the new network are the references of those in
-   the old one.
+  val get_subnetwork : ?copy:bool -> ?make_inputs:string array -> network -> string array -> network
+  (**
+   Constructs a subnetwork of nodes on which ``output_names`` depend,
```
mseri, line 678: I think it would be useful to name the options here:
```
``get_subnetwork ?copy ?make_inputs network output_names`` constructs a subnetwork of nodes of ``network`` ...
```

mreppen:
@mseri's comment makes sense to me as well I have a commit ready to push.

Another point I forgot to bring up: Would it perhaps make sense to drop or rename the `get_`-part of the function name? After all, it is not a getter function. Either plain `subnetwork` or something like `make_subnetwork`?

mseri:
Sounds reasonable to me to rename it, this could allow to keep backward compatibility although I don't see the need for this. I think we should wait to see what @jzstark @ryanrhymes or @tachukao think about it.

tachukao:
agree with the name change, looks good to me!

ryanrhymes:
Looks very good to me. Both `make_subnetwork` and `subnetwork` are fine by me.

mreppen:
@ryanrhymes, I picked `make_subnetwork`. @mseri's doc comment is also in the latest commit.

PR #517, opened 2020-04-14
Graph Slice node (resolves #483)

### LINKED ISSUE #483: Slicing node in Neural.Graph
I need to slice my input and feed the slices separately into the next step in my network. I could not find any interface to do the slicing, and
```
let open Owl.Neural.S in
Graph.(input [| 10; 10 |]
|> lambda (fun x -> Algodiff.Maths.get_slice [[1]] x))
```
yields
```
...
[ Node lambda_1 ]:
    Lambda       : in:[*,10,10] out:[*,10,10]
    customised f : t -> t
    prev:[input_0] next:[]
...
```
but the out dimension should be `1 x 10`.

Am I missing something?

mreppen:
A node producing a `get_slice` on its input.

Most code that is not boilerplate is for computing the output dimension. The calculation itself is pretty much the same as in `_enumerate_slice_def` (https://github.com/owlbarn/owl/blob/master/src/base/dense/owl_base_dense_ndarray_generic.ml#L99). The `assert` is a bit more aggressive (checks bounds) to avoid `Assert_failure ("src/base/core/owl_base_slicing.ml", 90, 10).` when running.

I am aware of special cases like that "later" axes do not need to be specified, and handle that. Are there others that I might have missed?

~~There is one potential extra feature: A bool flag to set whether trivial dimensions (those with length 1) should be remove, i.e., reshaped away. Is this too niche/specialized?~~

~~There is still a TODO for prettier (more detailed) printing.~~

Example:
```
# let nn = Neural.D.Graph.(input [| 5; 2 |] |> slice [[1; -1; 3]] |> get_network);;
val nn : Owl_neural.D.Graph.network =
  42878

[ Node input_0 ]:
    Input : in/out:[*,3,2]
    prev:[] next:[slice_1]

[ Node slice_1 ]:
    Slice : in:[*,3,2] out:[*,2,2]
    prev:[input_0] next:[]

# Neural.D.Graph.model nn (Arr.sequential [| 1; 5; 2 |]);;
- : Owl_algodiff_primal_ops.D.arr =

       C0 C1
R[0,0]  2  3
R[0,1]  8  9
```
```
# Arr.(sequential [|5;2|] |> get_slice [[1;-1;3]]);;
- : Arr.arr =
   C0 C1
R0  2  3
R1  8  9
```

### DIFF
--- a/src/base/misc/owl_utils_infer_shape.ml
+++ b/src/base/misc/owl_utils_infer_shape.ml
@@ -320,6 +320,40 @@ let split shape axis parts =
     parts
 
 
+let slice shape slice_spec =
+  let infer_len orig_len start stop ?step () =
+    let start = if start < 0 then orig_len + start else start in
+    let stop = if stop < 0 then orig_len + stop else stop in
+    let step =
+      match step with
+      | Some x -> x
+      | None   -> if start <= stop then 1 else -1
+    in
+    assert (
+      (start <= stop && step > 0 && stop < orig_len)
+      || (start > stop && step < 0 && start < orig_len));
+    let step_abs = abs step in
+    (abs (stop - start) + step_abs) / step_abs
+  in
+  let shape' =
+    List.mapi
+      (fun i slicei ->
+        let length = shape.(i) in
+        let infer_len_i = infer_len length in
+        match slicei with
+        | []                    -> length
+        | [ index ]             -> infer_len_i index index ()
+        | [ start; stop ]       -> infer_len_i start stop ()
+        | [ start; stop; step ] -> infer_len_i start stop ~step ()
+        | _                     -> failwith
+                                     "owl_utils_infer_shape: invalid slice specification")
+      slice_spec
+  in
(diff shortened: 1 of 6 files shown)


mreppen:
I added more detailed printing:
```
# ... slice [[1; -1; 3]; []; [0]] ...
...
[ Node slice_1 ]:
    Slice : in:[*,5,2,10] out:[*,2,2,1]
    Axes  : 0:[1; -1; 3] 1:[] 2:[0]
    prev:[input_0] next:[]
```

jzstark:
Cool! Thanks for this PR! Looks good to me. I'm just wondering if we can put the slicing shape inference routine to the [owl_utils_infer_shape.ml](https://github.com/owlbarn/owl/blob/master/src/base/misc/owl_utils_infer_shape.ml) file.

mreppen:
@jzstark I was not aware of that file, but that looks very appropriate. Latest commit moves it.

Issue #529, opened 2020-06-16
keepdims option for sum, std, sem, var, mean etc.


tachukao:
I often find myself writing code such as 
```
let mu = Arr.mean ~axis:0 x |> Arr.squeeze ~axis:[|axis|]
let stdev = Arr.std ~axis:0 x |> Arr.squeeze ~axis:[|0|]
```
It will be useful to have a `keepdims` option as in numpy [`sum`](https://numpy.org/doc/stable/reference/generated/numpy.sum.html), which automatically squeezes out the dimension we `sum` over if `keepdims` is set `true`.

In numpy, `keepdims` is true by default. For backward compatibility, it is probably best to make `false` the default of `keepdims` in Owl if we make these changes. Such changes should be straightforward to incorporate into Owl's Ndarray and Algodiff modules. After the proposed changes we will be able to write
```
let mu = Arr.mean ~axis:0 ~keepdims:false x
```

@ryanrhymes  @mseri  @ghennequin  @jzstark  thoughts?

mseri:
This sounds like a welcome change to me.

At this stage I think we should not be too worried about backward compatibility, and think more about having an ergonomic and consistent API before we draft a stable release.

I personally think `keepdims:true` is more convenient, but we need to test the repercussions of such change very carefully.

ghennequin:
a minor thing: `keepdims` should probably be `keep_dims` for general consistency with how distinct words are separated

ryanrhymes:
I agree, it looks like a very good addition to current function parameters. No need to be backward compatibility as Marcello said. :)

tachukao:
In my original post I completely messed up what I wanted to say.
The default of `keepdims` is `false` in numpy  (I said it's true in the original post) and for backward compatibility, it's best to have `keep_dims` as `true` in Owl. 
We squeeze the dimension if `keep_dims` is set to false (I said we squeeze it out if set to true in the original post).
I agree with @mseri that `keep_dims:true` is probably more convenient as will be backward compatible (unless that's not what you meant to say @mseri ?)

mseri:
I meant that backward compatibility is not a priority and I would not worry about that.

I find the most convenient default is the one that squeezes out the dimensions. Now I am not sure which one of them it is

tachukao:
[...]

PR #531, opened 2020-06-16
keep_dims option

tachukao:
As discussed in #529, I've added the `keep_dims` option for `sum`, `mean`, `var`, `std`, `sem`, `min`, `max`, `log_sum_exp`, `prod` and others.

The default option for `keep_dims` is `true` and is thus backward compatible.

@jzstark @ryanrhymes I'm not quite sure how `keep_dims` changes the computation graph in e.g., `owl_base_computation_operation.ml`. I've ignored the `keep_dims` option for now. Any ideas?

### DIFF
--- a/src/base/algodiff/owl_algodiff_ops.ml
+++ b/src/base/algodiff/owl_algodiff_ops.ml
@@ -523,16 +523,16 @@ module Make (Core : Owl_algodiff_core_sig.Sig) = struct
 
     and _sum =
       lazy
-        (fun ?axis ->
+        (fun ?axis ~keep_dims ->
           build_siso
             (module struct
               let label = "sum axis"
 
               let ff_f a = error_uniop label (pack_elt a)
 
-              let ff_arr a = Arr A.(sum ?axis a)
+              let ff_arr a = Arr A.(sum ?axis ~keep_dims a)
 
-              let df _cp _ap at = sum ?axis at
+              let df _cp _ap at = sum ?axis ~keep_dims at
 
               let dr a _cp ca =
                 match axis with
@@ -542,12 +542,16 @@ module Make (Core : Owl_algodiff_core_sig.Sig) = struct
                   let reps = Array.(make ndim 1) in
                   let axis = Owl_utils.adjust_index axis ndim in
                   reps.(axis) <- s.(axis);
-                  repeat !ca reps
+                  if keep_dims
+                  then repeat !ca reps
+                  else (
+                    s.(axis) <- 1;
+                    repeat (reshape !ca s) reps)
                 | None      -> !ca
             end : Siso))
 
 
-    and sum ?axis = Lazy.force _sum ?axis
+    and sum ?axis ?(keep_dims = false) = Lazy.force _sum ?axis ~keep_dims
 
     and _sum_reduce =
       lazy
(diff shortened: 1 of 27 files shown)


ryanrhymes:
I cannot immediately say whether or not it will impact thing in CGraph, do we have corresponding unit tests for cgraph? @jzstark

ryanrhymes:
@pvdhove Pierre can also comment if there is anything in his mind. Pierre did a lot of work on CGraph and is very familiar with it as well.

pvdhove:
From what I recall, here are a few changes that would need to be made if you want the extra argument to have the expected behaviour with CGraph (all files that I mention are in `owl/src/base/compute`) (at the moment, everything should compile but the argument will not have any impact if the functions are evaluated with CGraph):
- Adding `bool` arguments to the relevant functions in `owl_computation_type.ml`, and thus also in `owl_computation_symbol.ml`, `owl_computation_optimiser.ml`, and `owl_computation_cpu_init.ml`. Nothing else should change there (unless there is some optimisation to be done in the new case `keep_dims=false`, but it is not necessary and it is safer not to do any optimisation in the first place).
- Using the new parameter in the relevant functions in `owl_computation_operator.ml` and in `owl_computation_cpu_eval.ml`.
- If I understood correctly, the shape of the output depends on the `keep_dims` parameter; therefore, the function predicting the shape of the nodes should be changed in `owl_computation_shape.ml` -- I am not sure whether there is an existing `_infer_shape_xx` function that already returns the right shape with similar arguments.

For some more context about optimisation: what I had chosen in `owl_computation_cpu_init.ml` for the functions that you consider is *not* to be able to reuse the memory allocated to their inputs (called `parents` in the arguments, because the inputs of the functions correspond the parents in the computation graph). This behaviour is always safe even if you change the logic of the function, so nothing must be changed there. It might be possible to tell these functions that they can safely reuse the memory of their parents while they are computed for some extra memory optimisation, but that is very much implementation-dependent and a bit risky if changes are made later.

tachukao:
I have just pushed some changes that shall allow users to use the `keep_dims` option in CGraph.

There are several aspects of the code that I do not fully understand, and for `keep_dims=false` I have followed the example provided by `SumReduce`, which hopefully is the right thing to do.

To infer shape in `owl_computation_shape.ml` I have written a new helper function `_infer_shape_31`. This function calls `_infer_shape_04` (normally used by `sum`), when `keep_dims=true`. It calls `_infer_shape_10` (normally used by `sum_reduce`), when `keep_dims=false`.

In `owl_computation_cpu_evals.ml`, @pvdhove am I correct in my interpretation that `_eval_map_01` reuses the memory of the node by directly writing to it, and `_eval_map_00` does not? Since I did not implement the `keep_dims` option for functions such as `sum_`, `min_`, and `max_`, I am calling `_eval_map_00` when `keep_dims=true`. 

Also, do we have a unit test for the operations in CGraph?

pvdhove:
Everything looks good to me! Your interpretation of the ``_eval_map``s is right, it is simply to distinguish functions that are pure or impure (further memory optimisation is done in ``owl_computation_cpu_init.ml``). I haven't checked that the right shapes were computed because I got a bit lost in the discussion about the exact behaviour of ``keep_dims`` :-D but what you did seems reasonable.

For the optimisation part, I just don't understand why ``pattern_024`` is called when ``keep_dims=false``? It seems that the result is the same as ``pattern_000``, because ``pattern_024`` only does something special to ``SumReduce``. But that does not change anything, as long as the optimisation is propagated to the parent.

To my knowledge, there are no unit tests for CG, only examples (both in the ``examples/`` folder and in distinct repositories).

tachukao:
oh right, I forgot about the optimization part. I meant to ask. 

I didn't really understand what `pattern_024` and `pattern_000` do, and was simply coding by analogy. But now I gather `pattern_024` is an optimization special to `sum_reduce`. Is that right? 
And so in the latest commit, I have reverted back to `pattern_000` for `sum`, `min`, `max` etc, which I assume means no optimization. Is that right @pvdhove ?

In any case, we should probably open an issue to add unit tests for operations in `CG`.

mseri:
Thanks both for looking into it! I agree that a new issue for the unit tests would be good. @tachukao given that you are there, it may be a good idea to add some meta comments on why you are calling certain functions instead of other (e.g. the optimisation in pattern_xxx or _eval_map)

pvdhove:
> I didn't really understand what `pattern_024` and `pattern_000` do, and was simply coding by analogy. But now I gather `pattern_024` is an optimization special to `sum_reduce`. Is that right?

Yes, basically the optimization part performs a DFS of the computation graph before the evaluation of its nodes, and implements a few different techniques to make things more efficient: for instance, removing useless operations, pre-computing the parts of the graph that do not depend on the inputs (so that if the graph is evaluated multiple times with different inputs, those parts of the graph are only computed once), changing the order of the operations to make them faster or more precise, merging operations that can be performed in one go, etc. The ``pattern_024`` appears to replace ``SumReduce`` by ``Sum`` when ``axis`` has a certain shape, probably because the implementation is faster in this specific case.

> And so in the latest commit, I have reverted back to `pattern_000` for `sum`, `min`, `max` etc, which I assume means no optimization. Is that right @pvdhove ?

Yes!

> In any case, we should probably open an issue to add unit tests for operations in `CG`.

The various operations reuse the code written outside of the CGraph module, so I guess that is why unit tests with CG for specific operations have not seemed essential. However, it would be really good if there were unit tests for specific CG operations (the optimizations, creating the graph, manipulating the graph, etc.).

ryanrhymes:
As Pierre mentioned, yes the real operations of CG are provided by the module we plug in, so they need to be tested with a specific backend. As you all said, the generic graph operations shall be tested at least, totally agree.

tachukao:
@mseri  I've just add some comments. Since we aren't do any optimization for these operations, we are going with the defaults mostly `pattern_000` and `eval_map_00`

tachukao:
@ryanrhymes @mseri shall we merge this if there's nothing else that needs changing right away?

ryanrhymes:
The PR looks all good to me. Unit tests of CG is a separate issue we can take care of that in future. 

@mseri do you have further comments on this?

Issue #532, opened 2020-06-17
Subtraction in Algodiff


priamfive:
Hi,  I have been  experimenting with Algodiff and have had some odd behavior with subtraction. 
I'm not sure if I am doing something wrong. If I want to differentiate f(x)=1-x, I seem to get an odd result if I do it in the obvious way: 
`# #require "owl-top";;
# module AD = Owl.Algodiff.D;;
module AD = Owl.Algodiff.D
# let f x = AD.Maths.((F 1.0) - x);;
val f : AD.t -> AD.t = <fun>
# let f1 = AD.diff f;;
val f1 : AD.t -> AD.t = <fun>
# f1 (F 0.0);;
- : AD.t = [F(1)]
# let g x = AD.Maths.((F 1.0) + (neg x));;
val g : AD.t -> AD.t = <fun>
# let g1 = AD.diff g;;
val g1 : AD.t -> AD.t = <fun>
# g1 (F 0.0);;
- : AD.t = [F(-1)]
`
The second way (using g) works but the first (using f) has the wrong sign.  

Perhaps the second way is what is supposed to be used but this wasn't clear from the excellent draft Owl book.

tachukao:
@priamfive thanks for raising this issue. The culprit was this line 

https://github.com/owlbarn/owl/blob/5150b0aaae2d2cd79b4f4a099411879eaadcf417/src/base/algodiff/owl_algodiff_ops.ml#L998

Instead of `bt` it should have been `neg bt`. This is now fixed in the PR #533

priamfive:
Great - thanks for all the good work on Owl.

PR #533, opened 2020-06-17
fixed bug in sub forward mode

tachukao:
Fixed subtraction bug in forward mode raised in #532. There pointed to a _related_ bug in the forward mode of `lyapunov`.


ryanrhymes:
Calvin, thank you for fixing it.

PR #543, opened 2020-09-03
Fix access

mseri:
The following works:
```
let (.%[;..]) = Bigarray.Genarray.get;;
let sum x y = x.%[1;2;3] + y.%[1;2];;
```
While the following both fail with Error: Unbound value .%[]:
```
let (.%[;..]) = Bigarray.Genarray.get;;
let sum x y z = x.%[1;2;3] + y.%[1;2] + z.%[1];;
let sum x y z = x.%[1;2;3] + y.%[1;2] + z.%[1;];;
```
Defining both, e.g.
```
let (.%[;..]) = Bigarray.Genarray.get;;
let (.%[]) a v = Bigarray.Genarray.get a [|v|];;
let sum x y z = x.%[1;2;3] + y.%[1;2] + z.%[1];;
```
everything seems to work.

In owl_operator we were not defining the single accessor.

### DIFF
--- a/src/base/compute/owl_computation_optimiser.ml
+++ b/src/base/compute/owl_computation_optimiser.ml
@@ -74,9 +74,9 @@ module Make (Operator : Owl_computation_operator_sig.Sig) = struct
       | Asinh -> pattern_000 x
       | Acosh -> pattern_000 x
       | Atanh -> pattern_000 x
-      | Min (_keep_dims, _axis) -> pattern_000 x 
-      | Max (_keep_dims, _axis) -> pattern_000 x 
-      | Sum (_keep_dims, _axis) -> pattern_000 x 
+      | Min (_keep_dims, _axis) -> pattern_000 x
+      | Max (_keep_dims, _axis) -> pattern_000 x
+      | Sum (_keep_dims, _axis) -> pattern_000 x
       | SumReduce _axis -> pattern_024 x
       | Signum -> pattern_000 x
       | Sigmoid -> pattern_000 x
--- a/src/base/core/owl_operator.ml
+++ b/src/base/core/owl_operator.ml
@@ -150,9 +150,29 @@ end [@warning "-34"]
 module Make_Matrix (M : MatrixSig) = struct
   type ('a, 'b) op_t2 = ('a, 'b) M.t
 
-  let ( .%{;..} ) x i = M.get x i.(0) i.(1)
+  let ( .%{} ) x (i1, i2) = M.get x i1 i2
 
-  let ( .%{;..}<- ) x i = M.set x i.(0) i.(1)
+  let ( .%{;..} ) x i =
+    if Array.length i = 2
+    then M.get x i.(0) i.(1)
+    else
+      failwith
+        (".%{} requires two indices but "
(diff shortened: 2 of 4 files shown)


mseri:
In https://github.com/owlbarn/owl/pull/543/files#diff-94492bc03a4f468124d294fde74cf827L149 and the next one, should we add a check for Array.length = 2 ?

Octachron:
This is the reason why the [manual example](https://ocaml.org/manual/indexops.html#ss:multiindexing) defines the two kind of operators. Should I add an explicit warning?

mseri:
@Octachron that would be great. I was planning in sending a PR but if you have something in mind already, even better! 

I had read the documentation When I stumbled on this bug yesterday and completely missed it at first. Even though why it happens is clear now, the error maybe could be improved but I think a warning in the manual would already go a long way.

Octachron:
Errors or warnings in the compiler would need to rely on a generic mechanism, and not be specific to indexing operators. It sounds  more like a task for a configurable linter.

mseri:
I have also added the single-valued access to matrix element using a tuple and a check on the length of the array for matrix access (to avoid hidden surprises when using it)

jzstark:
Just checked that this PR works fine (all tests pass) on my computer with OCaml 4.11.0, but I'm still a bit confused about what specifically the problem this PR is trying to solve and its cause. A clearer description in detail would be really helpful for the reviewing. Thanks!

mseri:
Multi-index operators are restricted to index expressions that contain one or more semicolons ";", so in most cases when you define multi-index operators you also need to define conjointly a single index variant. For example, our operators do not work when you need to access an element in 1d arrays.

This PR defines the single index operators for NDarrays and adds a test to check that it is there for the future.

In doing this I have also noticred a weakness in the matrix multi index: You can give any number of indices and it would fail silently. So I both added a check on the number of inidces used and added a single index operator for matrices that takes a tuple, so guarantees that you give the correct number of indices. A new test checks that both operators give a coinciding result.

jzstark:
Cool! Thanks for the explanation. The fix made in this PR looks good to me. There is only small concern: currently the fixed syntax seem to allow multiple forms of indexing,  e.g. `x.%{0; 1}` equals to `x.%{0, 1}`, and I'm not quite sure if this is a design choice that fits with the rest of Owl. I think Liang @ryanrhymes  would be in a better position to judge this.

mseri:
It is only for matrices, which are special anyways, and if we have to choose I’d rather go with the triple rather than the array since it must have exactly two indices

mseri:
And to limit breakage I would mark the matrix extended access as deprecated

tachukao (approved):

ryanrhymes:
> Cool! Thanks for the explanation. The fix made in this PR looks good to me. There is only small concern: currently the fixed syntax seem to allow multiple forms of indexing, e.g. `x.%{0; 1}` equals to `x.%{0, 1}`, and I'm not quite sure if this is a design choice that fits with the rest of Owl. I think Liang @ryanrhymes would be in a better position to judge this.

This looks fine by me.

jzstark (approved):

PR #544, opened 2020-09-05
base/Optimize_sig: partial doc improvement

pveber:
While I was trying to understand the meaning of Adam learning rate update strategy parameters, I thought I could as well add what I'd found in the doc. I also fixed warnings related to {6 } and {7 } sectioning levels not being supported, and in any case one section per module is not optimal IMHO. I replaced them with module-attached descriptions, the result is better with odoc. Finally I changed code markup to `[ ... ]` which is the one supported by ocamldoc/odoc.

### DIFF
--- a/src/base/optimise/owl_optimise_generic_sig.ml
+++ b/src/base/optimise/owl_optimise_generic_sig.ml
@@ -8,29 +8,31 @@ module type Sig = sig
 
   open Algodiff
 
-  (** {6 Utils module} *)
-
+  (** Utils module *)
   module Utils : sig
     val sample_num : t -> int
     (** Return the total number of samples in passed in ndarray. *)
 
     val draw_samples : t -> t -> int -> t * t
     (**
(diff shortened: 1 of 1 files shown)


pveber:
I also fixed type related comments which were incorrectly placed

mseri:
I am adding @ryanrhymes as reviewer since this breaks the generation if documentation using rest.

Since the new odoc, which should be out very soon, supports latex on html output and does a rather nice job, I would be in favour of updating it (and to use the master branch of odoc to generate the docs). But this means that we would need to update the whole documentation.

pveber:
Sorry, looking at the `Makefile` I was under the impression that `odoc` was already the default. Of course if needed I can provide a patch without changing the code markup.

mseri:
The generation is external afaik. But I never touched it, so I am not 100% sure of its evolution. Let's do a patch without changing the markup for the moment. I will create a separate issue to discuss moving it to ocamldoc syntax

pveber:
here you go!

mseri (approved):

PR #549, opened 2020-10-22
Windows compatibility

kkirstein:
This is a first step towards Windows compatibility. Feel free to start merging or wait for the remaining issues to be fixed, depending on your release plans.
Currently, `Owl_maths.mulmod` and `Owl_maths.powmod` crash; other tests pass. I don't see any special reason for this, but compilation warnings give hints that there are symbol clashes "around" `math.h`. I will further investigate this.
You can find some docs and status in `INSTALL_WINDOWS.md`

### DIFF
--- a/src/base/misc/owl_io.ml
+++ b/src/base/misc/owl_io.ml
@@ -82,7 +82,7 @@ let mapi_lines_of_file f fname =
 (* similar to iteri_lines_of_file but for marshaled file *)
 let iteri_lines_of_marshal ?(verbose = true) f fname =
   let i = ref 0 in
-  let h = open_in fname in
+  let h = open_in_bin fname in
   Fun.protect
     (fun () ->
       let t1 = ref (Unix.gettimeofday ()) in
@@ -116,13 +116,13 @@ let mapi_lines_of_marshal f fname =
 
 (* save a marshalled object to a file *)
 let marshal_to_file ?(flags = []) x f =
-  let h = open_out f in
+  let h = open_out_bin f in
   Fun.protect (fun () -> Marshal.to_channel h x flags) ~finally:(fun () -> close_out h)
 
 
 (* load a marshalled object from a file *)
 let marshal_from_file f =
-  let h = open_in f in
+  let h = open_in_bin f in
   Fun.protect
     (fun () ->
       let s = really_input_string h (in_channel_length h) in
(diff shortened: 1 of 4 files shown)


mseri:
Thanks!

src/base/misc/owl_io.ml
```diff
@@ -82,7 +82,7 @@ let mapi_lines_of_file f fname =
 (* similar to iteri_lines_of_file but for marshaled file *)
 let iteri_lines_of_marshal ?(verbose = true) f fname =
   let i = ref 0 in
-  let h = open_in fname in
+  let h = open_in_bin fname in
```
mseri, line 85: Oh, well spotted!

mseri:
I would be in favour of merging progressively the patches. 
@jzstark @ryanrhymes do you have a strong opinion on this?

mseri (approved):

mseri:
@kkirstein can you update the appropriate opam files to add a lower bound on both the new `conf-openblas` and `"eigen" {>= 0.3.0}`

mseri:
Fantastic, thanks

src/base/misc/owl_io.ml
kkirstein: Yeah, just saw that by chance...

mseri:
I am going to merge these changes. I will make a new release once windows passes all the tests

jzstark:
Thanks to Kay-Uwe for investigating and helping with this issue.  Looks like the C macros I have implemented a long time ago needs some improvement on the Windows OS. 

However, I find that with the recent code, several examples such as `mnist_cnn.ml` generate error: `Exception: Failure "input_value: unknown custom block identifier".`  I don't think the problem is introduced in this PR.  Perhaps it's just my machine/version mismatch/bad network connection etc., but it does require some further investigation.  (Also IMO the `INSTALL_WINDOWS.md` should be combined into the `INSTALL.md` file, but that's just trivial matter.)

I will try to track down this problem when I'm available. But be it a bug or not, it does involve the issue of proper code reviewing.  Perhaps it goes unnoticed by some of us, but recently Liang and I have just added a set of project governance rules, involving the team structure, PR, making releases, etc. E.g. [one of them](https://github.com/owlbarn/owl#development) requires that a PR be reviewed by at least two team members.  (I was planning to check this PR tonight, and thought we can at least wait for a while on it :) 

Currently we are considering restructuring the team, since the working environment has changed a lot for both of us for now. And this governance is part of this management procedure. I do believe that it will be followed by the community from now on, so as to ensure the team can achieve a mutual understanding of the current project progress, and that this project can grow healthily in a controllable manner.

kkirstein:
I have identified the mentioned crash and added as issue #550 so we can track it.
@jzstark: Also for me it looks like there could be an issue with involved C macros, probably, there are different default definitions in `mingw`compared to "vanilla" `gcc`.

PR #551, opened 2020-10-24
Fix Dockerfile for master

ryanrhymes:
I simply copied docker/Dockerfile.ubuntu to overwrite the one in root directory, so we can fix docker hub build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,20 +19,23 @@ RUN ldconfig /opt/OpenBLAS/lib/
 ##################### PREREQUISITES ########################
 
 ENV DEBIAN_FRONTEND noninteractive
-RUN apt-get update && apt-get install -y pkg-config m4 zlib1g libplplot-dev plplot-driver-cairo
-RUN apt-get remove plplot-driver-qt -y
-RUN opam install alcotest utop
+RUN apt-get update -y
+RUN apt-get install -y m4 wget unzip aspcud libshp-dev libplplot-dev gfortran pkg-config git
+RUN cd /home/opam/opam-repository && git pull --quiet origin master
+RUN opam update -q
 
 ####################   INSTALL OWL  #######################
 
-COPY . owl
-WORKDIR /home/opam/owl
+ENV OWLPATH /home/opam/owl
 ENV OWL_DISABLE_LAPACKE_LINKING_FLAG 1
 ENV OWL_COMPILE_CFLAGS "-I/opt/OpenBLAS/include -I/home/opam/OpenBLAS/lapack-netlib/LAPACKE/include/ -L/opt/OpenBLAS/lib"
-RUN CFLAGS=${OWL_COMPILE_CFLAGS} opam pin .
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -19,20 +19,23 @@ RUN ldconfig /opt/OpenBLAS/lib/
 ##################### PREREQUISITES ########################
 
 ENV DEBIAN_FRONTEND noninteractive
-RUN apt-get update && apt-get install -y pkg-config m4 zlib1g libplplot-dev plplot-driver-cairo
-RUN apt-get remove plplot-driver-qt -y
-RUN opam install alcotest utop
+RUN apt-get update -y
+RUN apt-get install -y m4 wget unzip aspcud libshp-dev libplplot-dev gfortran pkg-config git
```
mseri, line 23: Note that `aspcud` is no longer a dependency for opam, can be removed
I also don't see the need for `unzip`, `wget` and `git`

In any case we can get rid of both lines and replace them with `opam depext owl owl-plplot` right after `opam update -q`: this will install all the necessary dependencies.

If the image uses ubuntu-20.04, then it does not even need the custom openblas (not the docker image nor the custom compiled library and flags).

Dockerfile
```diff
@@ -19,20 +19,23 @@ RUN ldconfig /opt/OpenBLAS/lib/
...
+RUN cd /home/opam/opam-repository && git pull --quiet origin master
+RUN opam update -q
 
 ####################   INSTALL OWL  #######################
 
-COPY . owl
-WORKDIR /home/opam/owl
+ENV OWLPATH /home/opam/owl
 ENV OWL_DISABLE_LAPACKE_LINKING_FLAG 1
 ENV OWL_COMPILE_CFLAGS "-I/opt/OpenBLAS/include -I/home/opam/OpenBLAS/lapack-netlib/LAPACKE/include/ -L/opt/OpenBLAS/lib"
```
mseri, line 31: I don't see why you still want to use this custom compilation, setting `PKG_CONFIG_PATH=/opt/OpenBLAS/pkgconfig` or wherever the custom openblas.pc is, should make it work fine

Dockerfile
```diff
@@ -19,20 +19,23 @@ RUN ldconfig /opt/OpenBLAS/lib/
...
-RUN CFLAGS=${OWL_COMPILE_CFLAGS} opam pin .
+RUN CFLAGS=${OWL_COMPILE_CFLAGS} opam install owl owl-top owl-plplot utop -y
```
mseri, line 32: Why also utop?

If you remove the pins, it will install the owl released on opam. I think this image should use owl from the master branch.

jzstark (approved):

Dockerfile
ryanrhymes: utop is included in the very begining, so people can have a friendly REPL.

ryanrhymes:
This file is exactly the same as Docker.ubuntu, it works fine and fixes the issue, so I will just keep it as it is. We can improve both if necessary in future.

PR #552, opened 2020-11-02
fix portability issue of 'long int'

### LINKED ISSUE #550: [win32] Owl_maths.mulmod fails on large inputs
On Windows `Owl_maths.mulmod` fails on large input numbers, e.g., close to `max_int`. Here is an excerpt of the respective unit test (run in `utop`):
```
#require "owl";;
let i = max_int;;
Owl_maths.mulmod (i-1) (i-1) 0;;
```
This should return `0`, but runs forever (probably hangs in the `while` loop of the respective C-code). Maybe it's an issue with integer bit size. Investigation is ongoing...

kkirstein:
This PR replaces `long` with `int64_t` from `<stdint.h>`. This improves portability, as according to C standard, `long` is only guaranteed to be at least 32 bit (it is 64 bit on Linux with gcc, but 32 bit with mingw on Windows). This PR fixes #550 and has been tested on Windows10/Cygwin with `x86_64-w64-mingw32-gcc` and with `gcc` on Ubuntu 20.04 (WSL2).

### DIFF
--- a/src/owl/core/owl_ndarray_contract.h
+++ b/src/owl/core/owl_ndarray_contract.h
@@ -10,22 +10,22 @@
 /* Define structure for contracting indices of a tensor. */
 
 struct contract_pair {
-  long dim;         // number of dimensions, x and y must be the same.
-  long dep;         // the depth of current recursion.
-  long drt;         // number of outer loops.
+  int64_t dim;         // number of dimensions, x and y must be the same.
+  int64_t dep;         // the depth of current recursion.
+  int64_t drt;         // number of outer loops.
   int64_t *n;       // number of iteration in each dimension, i.e. y's shape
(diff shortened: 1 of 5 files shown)


mseri (approved):
This looks more robust and safer than what we had before. Thank you very much!

ryanrhymes (approved):
Looks very solid. Thank you @kkirstein

mseri:
The failure is for the Lyapunov test, which was notoriously unstable. @tachukao do you think this can be merged?

ryanrhymes:
> The failure is for the Lyapunov test, which was notoriously unstable. @tachukao do you think this can be merged?

It happens sometimes ... seems fine now after I re-triggered the CI.

PR #554, opened 2020-11-05
Convolution bug fix

### LINKED ISSUE #477: Convolution implementation bug
Thanks to @ghennequin , we have find a bug in the implementation of convolution operations in `owl-core`. Here is a simple script to verify this error:

```ocaml
module N = Owl_dense_ndarray.S
(* module N = Owl_base_dense_ndarray_s *)
let inp = N.ones [|1; 5243; 1|]
let ker = N.ones [|100; 1; 1|]
let out = N.ones [|1; 5243; 1|]
let k' = N.conv1d_backward_kernel inp ker [|1|] out
```
Replacing the module N from `owl-core` to `owl-base` would lead to different results. More specifically, in the current code I use two implementations for convolution, and the second one is used only when the input and kernel sizes are large enough (decided according to a threshold value). 

As a temporary workaround, in the PR #478 I set the threshold value to max_int, therefore the first (and correct) algorithm is always used. Currently I'm working on fixing this issue.

jzstark:
This PR fixes #477.

### DIFF
--- a/src/owl/core/owl_ndarray_conv_impl.h
+++ b/src/owl/core/owl_ndarray_conv_impl.h
@@ -11,7 +11,7 @@
  * Code heavily inspired by Eigen (http://eigen.tuxfamily.org/).
  */
 
-#define IM2COL_THRESHOLD LONG_MAX // TODO: a temp hack to disable the second conv algorithm
+#define IM2COL_THRESHOLD 0 // 512 * 1024
 #define ALIGN_SIZE 32 // for AVX address alignment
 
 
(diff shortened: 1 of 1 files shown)


ryanrhymes (approved):

ryanrhymes:
@jzstark I don't know how many in the core team understand this part except you and me ... I think it is ok to merge.

mseri:
Cool. Would it be worth to add a test using the repro case on the issue?

jzstark:
@mseri Thanks for the suggestion! This test cases was already added in a previous PR.

PR #560, opened 2020-11-13
Revert some changes (ocamlformat.0.16.0-rc2)

gpetiot:
Hi, after some discussions we chose to lower the amount of changes impacting the `if` expressions. The last build was breaking after the `if` keyword whenever the following expression didn't fit the line and that wasn't too pleasing to the eye whenever the keyword was alone on its line.
I should have made it more obvious in the previous PR that the diff wasn't definitive, it still isn't as long as ocamlformat 0.16.0 is not available on opam. Sorry for the inconvenience!

Here is the diff of #556 + #560 : https://github.com/gpetiot/owl/compare/master...gpetiot:preview-ocamlformat.0.16.0


ryanrhymes (approved):
I think this looks better and more inttuitive :)

mor1 (approved):

jzstark (approved):

Issue #540, opened 2020-08-26
diff and grad don't complain when called on vector/matrix-valued functions


pveber:
I wonder what's the expected behavior for this code:
```ocaml
# open Owl
# let f x = Algodiff.D.Maths.cos x;;
val f : Owl_algodiff.D.t -> Owl_algodiff.D.t = <fun>
# let x = Algodiff.D.Mat.gaussian 3 3;;
val x : Owl_algodiff.D.t = [Arr(3,3)]
# Algodiff.D.grad f x |> Algodiff.D.Mat.print;;

          C0        C1        C2 
R0 -0.817393  0.311382  0.975356 
R1  0.701889 -0.859119  0.566911 
R2 -0.184544 -0.888166 -0.754372 
- : unit = ()
```
I had expected `grad` to complain that `f` evaluates to a matrix instead of a scalar. It doesn't seem to be bothered, and seems to be computing the matrix `\frac{ \partial f_{ij} }{ \partial x_{ij} }`, which is not really a gradient.

tachukao:
Yes we're currently not checking the shape of the output, and you're quire right we should probably do that.

PR #561, opened 2020-11-14
Fixed bug in aiso + checks for grad and diff + forward-mode derivative for split

tachukao:
1. We now check that the output of `grad` and the input of `diff` are scalars (as referenced in #540)
2. There was a somewhat obscure bug in building AISO operations in Algodiff. The bug occurs when the input to an `aiso` operation (e.g. `concatenate`) include reverse-mode and forward-mode with tags at different levels. I've a new unit test `nested_grad2` in `unit_algodiff_matrix_generic.ml` to explicitly test for this.
3. I've also implemented the forward-mode derivative of `split`, which required changing `SIAO`  in `Builder`

### DIFF
--- a/src/base/algodiff/owl_algodiff_generic.ml
+++ b/src/base/algodiff/owl_algodiff_generic.ml
@@ -48,6 +48,7 @@ module Make (A : Owl_types_ndarray_algodiff.Sig) = struct
 
   (* derivative of f (scalar -> scalar) at x, forward ad *)
   let diff' f x =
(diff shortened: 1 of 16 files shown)


mseri:
This PR contains a previous PR. I think it is easier to review after the other PR is merged. I don't know AISO enough to have an informed review of this.

tachukao:
I think the other PR has been merged

ryanrhymes (approved):

jzstark (approved):

tachukao:
shall we merge this now?

PR #562, opened 2020-11-21
Add initial FFT 2D implementation

jzstark:
This PR adds simple implementation of 2D fft and ifft operations, as required by the example in owl book.


ryanrhymes:
Very good! We can leave the further optimisation in future, sufficient for current book writing!

ryanrhymes (approved):

mseri (approved):
Looks good. Since you are touching the fft file, would you mind also replacing the asserts we have there with an explicative error?

jzstark:
Thanks for the reviewing. Personally I'm not very keen on replacing the assertions in a haste, but I will try to take a look at this issue later.

PR #567, opened 2021-01-03
fix random seed of algodiff unit test

tachukao:


ryanrhymes (approved):

jzstark (approved):
Cool! I'll try to submit this version to opam to see if it works.

mseri (commented):
Good as a temporary fix.

I think on the long run it would be better to have a way to maybe re-run these tests or run them X times and fail only if we detect too many failures of these runts.

mseri:
I want to ``approve'' it, for some reason github did not allow me to do it. So consider this the green tick

Issue #583, opened 2021-04-14
bug in `Arr.split`


ghennequin:
`Arr.split` fails with  `INDEX_OUT_OF_BOUNDS` in the following example:

```ocaml
let x = Arr.sequential [| 3; 1 |]
let s = Arr.split ~axis:1 [| 1 |] x
```

Since the index array (`[| 1 |]`) implies, in fact, no splitting, a more natural behaviour would be to return `[| x |]` itself in this case (or even better, `[| Arr.copy x |]`). Indeed, this actually behaves as intuitively expected:

```ocaml
let x = Arr.sequential [| 1; 3 |]
let s = Arr.split ~axis:0 [| 1 |] x
```

So `Arr.split` has inconsistent behaviour.

tachukao:
The bug is on line 8471 of `owl_dense_ndarray_generic.ml`

`let a = Owl_utils.adjust_index axis _d` 

should be 

`let a = Owl_utils.adjust_index axis x_dim`

I'll file a PR later this week.

ghennequin:
Wonderful, thanks @tachukao !

tachukao:
@ghennequin maybe we should close this one? I can't seem to close it

PR #584, opened 2021-04-29
fix bug in split

tachukao:
fixing bug #583


ryanrhymes (approved):
Thanks Calvin!

mseri (approved):
Is it worth adding a test?

tachukao:
I don't think we need a test for this one --- not sure where it would belong if we do have one. But happy to discuss

Issue #594, opened 2021-08-09
compiler warning 16 [unerasable-optional-argument]


kkirstein:
In OCaml `4.12.0` `warning 16 [unerasable-optional-argument]` has been widened: https://github.com/ocaml/ocaml/pull/9783, which leads to build failures (see below).
Fixing this would probably break the API, so I would suggest to silence these warnings with `[@@ocaml.warning "-unerasable-optional-argument"]`

Any comments on this?

```
File "src/base/dense/owl_base_dense_ndarray_generic.ml", line 432, characters 20-24:
Error (warning 16 [unerasable-optional-argument]): this optional argument cannot be erased.

File "src/base/dense/owl_base_dense_ndarray_generic.ml", line 432, characters 17-18:
Error (warning 16 [unerasable-optional-argument]): this optional argument cannot be erased.

File "src/base/dense/owl_base_dense_ndarray_generic.ml", line 483, characters 17-18:
Error (warning 16 [unerasable-optional-argument]): this optional argument cannot be erased.

File "src/base/dense/owl_base_dense_ndarray_generic.ml", line 483, characters 14-15:
Error (warning 16 [unerasable-optional-argument]): this optional argument cannot be erased.

File "src/base/dense/owl_base_dense_ndarray_generic.ml", line 509, characters 17-24:
Error (warning 16 [unerasable-optional-argument]): this optional argument cannot be erased.

File "src/base/dense/owl_base_dense_ndarray_generic.ml", line 535, characters 19-24:
Error (warning 16 [unerasable-optional-argument]): this optional argument cannot be erased.

File "src/base/dense/owl_base_dense_ndarray_generic.ml", line 535, characters 15-17:
Error (warning 16 [unerasable-optional-argument]): this optional argument cannot be erased.

File "src/owl/cblas/owl_cblas.ml", line 53, characters 58-62:
Error (warning 16 [unerasable-optional-argument]): this optional argument cannot be erased.

File "src/owl/cblas/owl_cblas.ml", line 53, characters 51-56:
Error (warning 16 [unerasable-optional-argument]): this optional argument cannot be erased.

File "src/owl/cblas/owl_cblas.ml", line 53, characters 40-48:
Error (warning 16 [unerasable-optional-argument]): this optional argument cannot be erased.          .
File "src/owl/cblas/owl_cblas.ml", line 53, characters 28-36:
Error (warning 16 [unerasable-optional-argument]): this optional argument cannot be erased.          .
[...]

PR #595, opened 2021-08-09
suppress warning 16

kkirstein:
This PR would fix issue https://github.com/owlbarn/owl/issues/594
It probably needs some discussion first, whether this is the right way to fix the mentioned compiler warnings.


jzstark:
Thanks! The fix looks OK to me. Still not quite clear about the reason we have these warnings, and currently I can't think of a better way to suppress them.  Any suggestion? @ryanrhymes

PR #606, opened 2022-04-12
Add support for ARM64

### LINKED ISSUE #597: Not able to install natively on Apple M1
I'm just trying Owl and couldn't get this to compile natively on a M1 Mac.
If you use the rosetta emulation on the Mac you can get a working system.

I used:
sudo arch -x86_64 before the rest of the "opam/master/shell/install.sh" command. Then opam just works for the full owl install. The other awkward thing is to put arch -x86_64 before all the brew installs eg plplot openblas etc.

mseri:
Thanks @jhgorse for the help!
Should fix #597, I tested this on an iMac M1 and my Macbook Pro, but it would be good if somebody else could test it as well.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -15,18 +15,18 @@ before compilation:
 - `OWL_CFLAGS` allows to change the default flags passed to the C targets,
   it defaults to
   ```
-  OWL_CFLAGS="-g -O3 -Ofast -march=native -mfpmath=sse -funroll-loops -ffast-math -DSFMT_MEXP=19937 -msse2 -fno-strict-aliasing -Wno-tautological-constant-out-of-range-compare"`
+  OWL_CFLAGS="-g -O3 -Ofast -funroll-loops -ffast-math -DSFMT_MEXP=19937 -msse2 -fno-strict-aliasing"`
   ```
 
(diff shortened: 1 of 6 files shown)


mseri:
A parallel PR is needed for `eigen`, to remove `-march=native` or to detect it during build like we are doing here

ryanrhymes:
Awesome, looks very good to me.

ryanrhymes (approved):

ryanrhymes:
@jzstark can you also test on your mac?

xinslu:
Sorry, I wasn't pinged but I tried it out and it works on my MacBook Air M1 2020.

ddph3dg:
I still can't get this install to work. I needed to set the EIGEN_FLAGS to get eigen installed but fail with a ld error on owl:
ld: file not found: @rpath/libgcc_s.1.1.dylib for architecture arm64
It seems everything compiles OK but there is a hardcoded reference. A quick search seems to indicate openBlas gets compiled with this somehow.

mseri:
Having to set `EIGEN_FLAGS` is expected, the architecture detection has not been ported to eigen yet.

The error you get seems related to a mix of rosetta x86 libraries and arm64 in the system, `@rpath` is the runpath for macos dynamic linker.

PR #619, opened 2022-06-06
unpack_arr would fail in nested forward/reverse mode

ghennequin:
This is fixed by first extracting the `primal'` value.


mseri (approved):

tachukao:
If we use primal', I don't think we need unpack_arr?

ghennequin:
We do need `unpack_arr` (primal' still returns a `Algodiff.D.t` value).

tachukao (approved):
Looks good!

PR #624, opened 2022-10-12
Add the NeNormal initializer

samoht:
See https://www.tensorflow.org/api_docs/python/tf/keras/initializers/HeNormal


samoht:
Note: I don't know how critical for performance this function is for - if it is, it might make sense to move the computation of `r0`, ..., `r3` after the pattern-matching.

PR #630, opened 2023-01-21
Move to OCaml 5.0; removing zoo and aeos

jzstark:
- move to OCaml 5.0 by changing a line in the dune file in `base` 
- removing the `zoo` module  
- remove the `aeos` module 

Currently the `make test` still returns error message on my machine:

```
File "test/dune", line 2, characters 8-19:
2 |  (names test_runner)
            ^^^^^^^^^^^
(cd _build/default/test && ./test_runner.exe)
Command got signal SEGV.
```

EDIT:

I checked the test cases, and found that as long as I remove the following two test sets, `make test` works fine:

```
"linear algebra", Unit_linalg.test_set second test
("algodiff matrix", Unit_algodiff_matrix.[ Reverse.test; Forward.test ])
```


ryanrhymes:
The `make test` passed on my machine, Win10 Enterprise, Ubuntu on WSL. What is your environment?

Issue #635, opened 2023-03-12
`libgfortran.so` not linked and causing compilation error


jcreinhold:
Thank you very much for the great work with this package.

I'm trying to build this package on a CentOS7 machine with `gcc` v11. I have an OpenBLAS installation compiled from source and a corresponding shared library object whose directory is linked to in the `OWL_FLAGS` which I have changed and provided below (with some of the directories replaces with `...` for space).

However, I'm receiving the following error:

```
#=== ERROR while compiling owl.1.1 ============================================#
# context              2.1.3 | linux/x86_64 | ocaml-option-flambda.1 ocaml-option-fp.1 ocaml-option-nnp.1 ocaml-variants.4.14.0+options | file:///.../opam-repository#25bba47d
# path                 .../.opam-switch/build/owl.1.1
# command              .../bin/dune build -p owl -j 79
# exit-code            1
# env-file             .../log/owl-16623-12e153.env
# output-file          .../log/owl-16623-12e153.out
### output ###
# [...]
# -> stdout:
# -> stderr:
#  | .../bin/ld: warning: libgfortran.so.5, needed by .../libopenblas.so, not found (try using -rpath or -rpath-link)
#  | .../libopenblas.so: undefined reference to `_gfortran_etime@GFORTRAN_8'
#  | .../libopenblas.so: undefined reference to `_gfortran_concat_string@GFORTRAN_8'
#  | collect2: error: ld returned 1 exit status
# Fatal error: exception Failure("Unable to link against openblas.")
# Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
# Called from Dune__exe__Configure.(fun) in file "src/owl/config/configure.ml", line 223, characters 8-51
# Called from Configurator__V1.main in file "otherlibs/configurator/src/v1.ml", line 734, characters 4-7
# Re-raised at Configurator__V1.main in file "otherlibs/configurator/src/v1.ml", line 742, characters 11-42
# Called from Dune__exe__Configure in file "src/owl/config/configure.ml", line 186, characters 2-1023
```

Here are the `OWL_FLAGS`:

```
[...]

Issue #625, opened 2022-11-04
Painful installation on MacOS


jonathan-laurent:
I tried to install owl on MacOS Monterey (x86-64) after running `brew install openblas`.
Ideally, `opam install owl` should just work. However, I had to do:

```
export PKG_CONFIG_PATH="/usr/local/opt/openblas/lib/pkgconfig"
export LDFLAGS="-L/usr/local/opt/openblas/lib -L/usr/local/Cellar/gcc/12.2.0/lib/gcc/12/"
export CPPFLAGS="-I/usr/local/opt/openblas/include"

ln -s /usr/local/Cellar/gcc/12.2.0/lib/gcc/12/libquadmath.0.dylib /usr/local/lib/libquadmath.0.dylib
ln -s /usr/local/Cellar/gcc/12.2.0/lib/gcc/12/libgcc_s.1.1.dylib /usr/local/lib/libgcc_s.1.1.dylib

opam install owl
```

The first block since to be necessary since homebrew's openblas installation is kegonly (since MacOS ships with its own version of openblas). Removing it results in the following error:

```
#=== ERROR while compiling conf-openblas.0.2.1 ==============================#
# context     2.0.6 | macos/x86_64 | ocaml-base-compiler.4.14.0 | https://opam.ocaml.org#67c37f4f
# path        ~/.opam/research/.opam-switch/build/conf-openblas.0.2.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build sh -exc cc $CFLAGS $(pkg-config --cflags openblas) test.c $(pkg-config --libs openblas)
# exit-code   1
# env-file    ~/.opam/log/conf-openblas-67509-a00279.env
# output-file ~/.opam/log/conf-openblas-67509-a00279.out
### output ###
# [...]
# to the PKG_CONFIG_PATH environment variable
# No package 'openblas' found
# ++ pkg-config --libs openblas
# Package openblas was not found in the pkg-config search path.
# Perhaps you should add the directory containing `openblas.pc'
# to the PKG_CONFIG_PATH environment variable
# No package 'openblas' found
# + cc test.c
# test.c:2:10: fatal error: 'cblas.h' file not found
# #include <cblas.h>
#          ^~~~~~~~~
# 1 error generated.
```

Removing the second block results in the following error:

```
#=== ERROR while compiling owl.1.0.2-1 ======================================#
[...]

PR #609, opened 2022-04-14
Support Arm64 (and apple M1)

mseri:
Resubmitting as requested

to try this PR on mac use
```
opam pin -n git+https://github.com/mseri/owl.git#arm64 --with-version=1.1.0
PKG_CONFIG_PATH="/opt/homebrew/opt/openblas/lib/pkgconfig" opam install owl.1.1.0
```

PR #636, opened 2023-03-13
add additional environment variables and add libraries/flags to linki…

jcreinhold:
RFC

This PR addresses the problems raised in #635

For non-standard build setups, this adds `OWL_LDFLAGS`, `OWL_CPPFLAGS`, and `OWL_LDLIBS` which correspond to the [usual `gcc` flags](https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html#index-LDLIBS).

I personally don't think these are excessive, but we could change this to only add `OWL_LDFLAGS` (or, e.g., `OWL_LIBRARY_FLAGS`) which would be sufficient if this PR adds too many new environment variables. The current implementation in this PR, however, allows the user to specify flags like they would if they were directly using `gcc`. Even though the flags are eventually combined into `c_flags` and `c_library_flags`, this allows users to not cram everything into something labeled `OWL_CFLAGS` and (the minimally required for this to build) `OWL_LDFLAGS`.

This also adds the generated flags to the OpenBLAS linking tests with `c_test` which were previously failing for the reasons mentioned in #635. This change is required for non-standard builds; the linking test will fail simply because it's lacking the correct flags—even though `owl` would build with `OWL_CFLAGS` and something akin to `OWL_LDFLAGS`.

### DIFF
--- a/src/owl/config/configure.ml
+++ b/src/owl/config/configure.ml
@@ -76,13 +76,15 @@ let get_ocaml_devmode_flags _c =
   if not enable_devmode then [] else [ "-w"; "-32-27-6-37-3" ]
 
 
+let clean_env_var env_var =
+  Sys.getenv env_var
+  |> String.trim
+  |> String.split_on_char ' '
+  |> List.filter (fun s -> String.trim s <> "")
+
+
 let default_cflags =
-  try
-    Sys.getenv "OWL_CFLAGS"
-    |> String.trim
-    |> String.split_on_char ' '
-    |> List.filter (fun s -> String.trim s <> "")
-  with
+  try clean_env_var "OWL_CFLAGS" with
   | Not_found ->
     [ (* Basic optimisation *)
       "-g"
@@ -102,7 +104,20 @@ let default_cflags =
     ]
 
 
-let default_libs = [ "-lm" ]
+let default_cppflags =
+  try clean_env_var "OWL_CPPFLAGS" with
+  | Not_found -> []
+
+
+let default_ldflags =
+  try clean_env_var "OWL_LDFLAGS" with
+  | Not_found -> []
+
+
+let default_ldlibs =
+  try clean_env_var "OWL_LDLIBS" with
+  | Not_found -> [ "-lm" ]
+
 
 let get_expmode_cflags _c =
   let enable_expmode = bgetenv "OWL_ENABLE_EXPMODE" in
@@ -197,12 +212,30 @@ let () =
           ~default:openblas_default
           (C.Pkg_config.get c >>= C.Pkg_config.query ~package:"openblas")
       in
-      if not
-         @@ C.c_test
-              c
-              test_linking
-              ~c_flags:openblas_conf.cflags
-              ~link_flags:openblas_conf.libs
+      let openmp_config = get_openmp_config c in
+      (* configure link options *)
+      let libs =
+        []
+        @ default_ldflags
+        @ default_ldlibs
+        @ openblas_conf.libs
+        @ cblas_conf.libs
+        @ default_gcc_path
+        @ get_accelerate_libs c
+        @ openmp_config.libs
+      in
+      (* configure compile options *)
+      let cflags =
+        []
+        @ default_cflags
+        @ default_cppflags
+        @ openblas_conf.cflags
+        @ cblas_conf.cflags
+        @ get_devmode_cflags c
+        @ get_expmode_cflags c
+        @ openmp_config.cflags
+      in
+      if not @@ C.c_test c test_linking ~c_flags:cflags ~link_flags:libs
       then (
         Printf.printf
           {|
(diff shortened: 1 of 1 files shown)


jcreinhold:
FWIW, this solves many of the install problems (really linking problems) that occur with this library; albeit, this places the responsibility on the user to correctly configure the library. However, we're implicitly relying on them doing this by asking them to `ln -s` certain libraries to certain locations so they can be found by the compiler.

For example, I encountered problems linking `libquadmath.dylib` on my Macbook when I was installing the library. I am using a Apple M1 processor, so I incorporated the modifications to `owl_core_utils.c` and `owl_macros.h` listed in [this PR](https://github.com/owlbarn/owl/pull/609)—keeping everything else as in this current PR.

Instead of doing ad hoc `ln -s` or other work arounds that are particularly fragile, I was able to add `OWL_LDFLAGS="-L/Users/jcreinhold/homebrew/Cellar/gcc/12.2.0/lib/gcc/current"` and `OWL_LDLIBS="-lquadmath"` and successfully installed `owl` (building from source).

I was also able to use this PR (and @mseri `aarch64` fix) to build this package at work on a very painful build system.

While it's a bit of a burden for the user to learn about how to configure this, I could add in some tutorial material, and perhaps an install FAQ based on the install issues, and the future install problems can be reduced to figuring out the right combination of flags to pass via these proposed environment variables.

nilsbecker:
I would very much welcome such a PR. I am used to setting linking and C flag variables for installing lacaml and gsl, and it works well there to use homebrew installed libs after reading some basic instructions in the README.
(Currently, not sure how to get owl running on my m1 mac. Can I opam pin some fork of owl to get this PR + the relevant bits of the arm64 PR?)

mseri (approved):

mseri:
I included the change both in my arm64 and arm64-bak branch

nilsbecker:
i just tied https://github.com/owlbarn/owl/issues/625#issuecomment-1312197649 and that worked for me -- i guess it's using your updated arm64 branch yes?

mseri:
Yes

nilsbecker:
On another M1 mac I tried the same but keep getting a linker error that says `ld: library 'lapacke' not found`. any hints?

oemmerson:
@mseri How come the changes in your fork haven't been merged?

mseri:
> On another M1 mac I tried the same but keep getting a linker error that says `ld: library 'lapacke' not found`. any hints? `brew install lapack` did not help so far. It's confusing as I'm not aware of relevant differences between the two computers.

I don't know. You can check if `OWL_DISABLE_LAPACKE_LINKING_FLAG=1` fixes it

nilsbecker:
> OWL_DISABLE_LAPACKE_LINKING_FLAG=1

indeed that did allow it to install! the working command line was:
`PKG_CONFIG_PATH="/opt/homebrew/opt/openblas/lib/pkgconfig" OWL_DISABLE_LAPACKE_LINKING_FLAG=1 opam install owl.1.1.0`

i.e. without adding a path to the separately installed lapack library. no idea why it was necessary on this machine and not the other one.

nilsbecker:
ok, while disabling the lapacke linking did allow owl to get installed, it then does not work properly in an application. i get loads of linking errors saying that some lapacke symbols are not found. 

adding the brew-installed lapack to the `PKG_CONFIG_PATH` when installing owl does not work either. so still no luck overall

mseri:
I don't know, indeed `OWL_DISABLE_LAPACKE_LINKING_FLAG=1` should not exist imho. If pkg-config has reasons to link lapacke it means it should be linked, but it was worth a try. Not sure what happens, do you have a mix of x86 and arm pagackes installed via homebrew? That can mess it up for example

nilsbecker:
no that should not be the problem. i did a rebuild of the homebrew installation from an exported package list. 

i just noticed that i have an env variable set: 
`OWL_MANUAL_LDFLAGS=-L/opt/homebrew/opt/gfortran/lib/gcc/current -L/opt/homebrew/opt/openblas/lib`
which i define in my bash_profile. i don't remember where this came from (shame on me). is this something recognized by owl? should i unset it?

also my default `PKG_CONFIG_PATH=/opt/homebrew/opt/openblas/lib/pkgconfig:/opt/X11/lib/pkgconfig:/usr/X11/lib/pkgconfig`...

nilsbecker:
i think the manual linking flags were from a previous workaround for installing owl. i just retried with `LDFLAGS=$OWL_MANUAL_LDFLAGS opam install owl.1.1.0` but that did not remove the lapacke linking error, alas.

nilsbecker:
Ok, I've got it now. Mea culpa. For a reason I cannot fathom, my full opam switch was an x86_64 binary without me knowing it. It silently ran under Rosetta2 but when trying to link arm64-only libraries, that did not work anymore somehow. So @mseri 's initial hunch was correct: some x86 cruft was still around: all of the switch! 

I was really convinced that this was a new install of ocaml via opam and not an old file tree transferred from a backup, but now i'm not so sure anymore.

Bottom line: the intallation instructions referenced above, pinning @mseri 's branch do work for me now.

jzstark:
Close for now since the configuration has been updated.

PR #658, opened 2024-03-06
Enable github actions for testing

mseri:
This tests separately the three packages, should catch packaging issues before releases are made. It also tests the arm builds for macos (they will fail until https://github.com/owlbarn/owl/pull/609 is merged)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,51 @@
+name: Main workflow
+
+on:
+  - pull_request
+  - push
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-14 # beta runner for macos arm64
+          - macos-latest
+          - ubuntu-latest
+          # - windows-latest
+        ocaml-version:
+          - 5.1.x
+          - 4.14.x
+          - 4.12.x
+          - 4.10.x
+
+    runs-on: ${{ matrix.os }}
+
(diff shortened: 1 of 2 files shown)


jzstark:
Thanks a lot for adding test for various platforms other than Ubuntu. The test shows that building on most macos systems fails. Besides the main workflow based on `ubuntu-latest`, I have also tested a Github action on `macos-latest`, where Owl builds well. Perhaps this [mac.yml](https://github.com/owlbarn/owl/blob/main/.github/workflows/mac.yml) would be helpful. I believe in the tests we should build Owl from source instead of `opam`  repository, where the Owl code is actually outdated.

mseri:
The tests here are building the code from the repository not from the released packages in opam. The packages are automatically pinned to the version in the PR branches by the setup-ocaml action (in the way it is configured now) and then use the pinned version. 

I will have a look at the Mac action you linked, the failure on Mac arm is expected but I was not expecting the other failure

mseri:
I looked at the mac script, it works only because it is restricted to ocaml 5.1, I think if you enable 4.14 or earlier you should see the same failures.

The main advantage of using setup-ocaml in the way suggested here is that it will ensure that the depext, opam files and tests are properly setup for release and should work properly for the users that want to use the dev version.

jzstark:
True, this issue arm64 issue is one that requires fixing. Thanks a lot for your help, and I'll take a  look at both PRs soon.

PR #676, opened 2024-07-21
Add Nix/Flake support

yogeshsajanikar:
- Use tweag/opam-nix for supporting and building Owl
- Add 'owl' as a default library to build
- `nix build` builds owl library
- `nix build .#owl-base .#owl .#owl-top` to build all the components

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,11 @@ setup.log
 
 # editor config files
 .vscode
+
+# Ignore direnv files
+.envrc
+.direnv/
+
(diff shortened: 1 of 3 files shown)


yogeshsajanikar:
Add the nix support for owl. The next step is to add these packages to `ocamlPackages` in nixpkgs. This way it should be possible to run `nix run nixpkgs#utop` with `owl-top`. I am currently trying to use overlay to make this work. Once this PR is approved or merged, it should be possible to modify nixpkgs.

PR #682, opened 2024-11-19
.os.yml: drop support for 4.12, check also 5.2

mseri:
And later we could switch to 5.3, to ensure compatibility with recent compilers


jzstark:
After checking the supporting distribution version on different os's ([link](https://repology.org/project/ocaml/versions)), and the fact that the current Long Term Support release of OCaml is 4.14, I think it is safe to drop support to 4.10 now. Thanks @mseri !

